PR #4, opened 2021-01-19
Generate DID

danlee1025:
- Added method to generate DIDs according to [sidetree](https://identity.foundation/sidetree/spec/#create) spec.
- Added method to resolve long form DIDs and return initial state described in [spec](https://identity.foundation/sidetree/spec/#long-form-did-uris)
- Added `Keys` module to generate signing/encryption keys and convert to jwk.
- Reece added an extension to the json canonicalization library to work on maps.

Run `rake healthcards:did` to generate a random did. The task will print out the short, long form, and the payload.


Rakefile
```diff
@@ -16,3 +17,19 @@ task :rubocop do
...
+namespace :healthcards do
+  desc 'Generates a random DID and prints out payload.'
+  task :did do
+    include Dids
+    include Keys
+    enc_key = Keys.generate_key
+    sign_key = Keys.generate_key
+    update_key = Keys.generate_key
+    rec_key = Keys.generate_key
+    did = Dids.generate_did(enc_key[:publicJwk], sign_key[:publicJwk], update_key[:publicJwk], rec_key[:publicJwk])
```
radamson, line 30: Typically when including a module you'll just reference the module methods directly without prepending the module. In a regular class something like this shouldn't work, unless the methods are defined as class methods (not sure why it works here though...maybe rake does something special that I don't understand)
```suggestion
    enc_key = generate_key
    sign_key = generate_key
    update_key = generate_key
    rec_key = generate_key
    did = generate_did(enc_key[:publicJwk], sign_key[:publicJwk], update_key[:publicJwk], rec_key[:publicJwk])
```

Rakefile
```diff
@@ -16,3 +17,19 @@ task :rubocop do
...
+    puts did[:didShort]
+    puts did[:didLong]
+    puts JSON.pretty_generate(Dids.resolve_did_long(did[:didLong]))
```
radamson, line 33:
```suggestion
    puts JSON.pretty_generate(resolve_did_long(did[:didLong]))
```

lib/health_cards/dids.rb
```diff
@@ -0,0 +1,81 @@
...
+    long_payload_encoded = encode long_payload.to_json_c14n
+    {
+      didShort: "did:ion:#{suffix}",
+      didLong: "did:ion:#{suffix}:#{long_payload_encoded}"
+    }
+  end
+
+  def resolve_did_long(did_long)
+    # did format: did:ion:<did-suffix>:<long-form-encoded-data>
+    encoded_payload = did_long.split(':')[-1]
```
radamson, line 78:
```suggestion
    encoded_payload = did_long.split(':').last
```

radamson:
I'm starting to rethink whether it's worth merging this now. I'm leaning towards just closing this and we can reopen it once we need the functionality. Since DIDs were removed from the spec this is really just another piece of code to maintain.

We should still keep or adapt the useful parts related to generating the JWK though since that's what's used in the spec. Could do that in a new PR for a cleaner break and to make revisiting this simpler in the future.

Thoughts @danlee1025?

PR #7, opened 2021-02-16
WIP: Save credential as QR Code

danlee1025:
Added `save_as_qrcode` function to save credential.

Run `rake "healthcards:vc[payload]"` to generate vc, `credential.png`, QR code containing the credential as payload, will get created.


lib/health_cards/verifiable_credential.rb
```diff
@@ -44,6 +47,14 @@ def credential
 
     def jwt; end
 
+    def save_as_qrcode
```
radamson, line 50: Can we add a filepath argument here and then have `QR_CODE_PATH` be the default?

lib/health_cards/verifiable_credential.rb
```diff
@@ -44,6 +47,14 @@ def credential
...
+      qrcode = RQRCode::QRCode.new(credential.to_json, size: 40)
```
radamson, line 51: I think we need to set `mode: number` too to make it a numeric QR Code.

[Every Health Card can be embedded in a QR Code](https://smarthealth.cards/#numerical-encoding) section of the spec has some details on the QR code settings that we should double check.

danlee1025: Thanks, I did not see this QR code section.

radamson: It's a pretty new section so I don't think it was actually around when we originally created this task.

PR #8, opened 2021-03-02
Patient UI with Getting and posting data

pmadria:
This creates a patient UI form that is able to take and submit data. Run `npm start`to start the react server for the form and run `bundle exec ruby app.rb` to obtain information from the form in a get or post data about the patient in another server.

Localhost:3000 => react server
Localhost:8000 => localhost server for patient 

Sample GET request: http://localhost:8000/Patient/ for Postman or whatever tool you use.


frontend/src/PatientForm.js
```diff
@@ -0,0 +1,119 @@
...
+        <input
+          type="text"
+          value={email}
+          onChange={(event) => setEmail(event.target.value)}
+        />
+      </label>
+      <label>
+        Date of Birth:
+        <input
+          type="text"
```
Jammjammjamm, line 109: I think you want `date` rather than `text` here.

app.rb
```diff
@@ -0,0 +1,44 @@
...
+
+require 'sinatra'
+require 'json'
+
+set :port, 8000
+set :static, true
+set :public_folder, 'static'
+set :views, 'frontend/src'
+
+post '/Patient/' do
```
Jammjammjamm, line 11:
```suggestion
post '/Patient' do
```

app.rb
```diff
@@ -0,0 +1,44 @@
...
+  @@mi = params['name'][0]['given'][1]
+  @@ln = params['name'][0]['family']
+  @@suffix = params['name'][0]['suffix']
+  @@gender = params['gender']
+  @@telecom = params['telecom'][0]['phone']
+  @@email = params['telecom'][0]['email']
+  @@birth_date = params['birthDate']
+end
+
+get '/Patient/' do
```
Jammjammjamm, line 25:
```suggestion
get '/Patient' do
```

Jammjammjamm (commented):
[Foreman](https://github.com/ddollar/foreman) can handle this situation where you need to run ruby & js:

- `gem install foreman`
- create a file called `Procfile.dev` that looks something like:
```
api: bundle exec ruby app.rb
client: npm start
```
- run `bundle exec foreman start -f Procfile.dev` (or better yet, make a `bin/serve` script to do this)


Is this what it is supposed to look like?
Screen Shot 2021-03-04 at 12 16 29 PM

ssayer:
Closing this due to the move to Rails, but leaving the branch in case we want to refer to the JS code later

PR #10, opened 2021-03-04
DVCI-40: Code coverage

Jammjammjamm:
Adds and configures simplecov. Also some minor updates to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,11 @@
-/.bundle/
-/.yardoc
-/_yardoc/
-/coverage/
-/doc/
-/pkg/
-/spec/reports/
-/tmp/
+.bundle
+.yardoc
+_yardoc
+coverage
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,11 +1,11 @@
...
-/spec/reports/
-/tmp/
+.bundle
+.yardoc
+_yardoc
+coverage
+doc
+pkg
+tmp
+node_modules
```
radamson, line 8: Why have this ignore both matching directories and files, instead of just directories? The GitHub `.gitignore` templates also have them with the slashes.

This is a nit, but I don't think it needs to change.

radamson (commented):
Tested it out and it looks good. One minor nit on the .gitignore

radamson (approved):

PR #15, opened 2021-03-12
Add Configurable well-known endpoint

ssayer:


config/well-known.yml
```diff
@@ -0,0 +1,14 @@
+default: &default
+  authorization_endpoint: https://ehr.example.com/auth/authorize
```
Jammjammjamm, line 2: I'd prefer to see the endpoints configured via environment variables rather than having to commit each environment's configuration.

Jammjammjamm (approved):

PR #16, opened 2021-03-15
Add Immunizations Support

ssayer:
This PR adds the ability to track immunization of patients through the web UI. The FHIR mapping code from Patient has been refactored into an ActiveRecord concern for reuse. The FHIR Immunization resource is a composite of the Immunization and Vaccine models. These were separated to make it easy to add new vaccines as they become available (e.g. AstraZeneca). Vaccine data was taken from CDC website.


db/schema.rb
```diff
@@ -10,12 +10,31 @@
...
     t.string "json"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
   end
 
+  create_table "vaccines", force: :cascade do |t|
+    t.string "code"
+    t.string "name"
+    t.integer "doses_required"
+    t.integer "days_to_effectiveness_from_last_dose"
```
Jammjammjamm, line 35: I don't get this column when I run the migrations.

app/helpers/patients_helper.rb
```diff
@@ -1,4 +1,7 @@
 # frozen_string_literal: true
 
 module PatientsHelper
+  def display(attribute)
+    attribute
```
Jammjammjamm, line 5: What is up with this method?

db/schema.rb
ssayer: I was considering using that right when I started, but decided against it. I must have run the migration before deleting it. I'll remove it from the schema file.

app/helpers/patients_helper.rb
ssayer: I originally had it spitting out an html fragment. It was just a few tags, so a partial would have been more text than just copy/pasting it, so I decided to create a helper method. Rubocop didn't like me using raw() in the helper so I got rid of it and the html and just forgot to remove display(). I'll remove it since you're redesigning the UI anyways.

Jammjammjamm:
I get this error when creating an immunization:

Screen Shot 2021-03-17 at 8 21 55 AM

ssayer:
> I get this error when creating an immunization:
> 
> Screen Shot 2021-03-17 at 8 21 55 AM
> 
> FYI I plan on replacing this input with something like this because the ux for this input is pretty bad: https://goiabada.blog/one-datetime-two-inputs-doing-it-in-rails-c538f628ec9e

This should be fixed now. You may have to re-create your database. The serialization code was overriding the Rails magic that converts UI inputs to Dates etc. but I figured out the trick to do it correctly.

Jammjammjamm:
Still getting the same error.

Jammjammjamm (approved):

PR #17, opened 2021-03-17
Added jwks endpoint

danlee1025:
Added `/.well-known/jwks.json` endpoint.


app/controllers/well_known_controller.rb
```diff
@@ -1,8 +1,15 @@
...
+require_relative '../../lib/health_cards'
+
 # WellKnownController exposes the .well-known configuration to identify relevant urls and server capabilities
 class WellKnownController < ApplicationController
   def index
     render json: Rails.configuration.well_known.to_json
   end
+
+  def jwks
+    @issuer = HealthCards::Issuer.new ::Configuration.key_path
```
ssayer, line 12: We need to figure out a consistent way to do our config loading. My preference would be to put things in the existing Rails app config (see index method) rather than creating our own config object.

danlee1025: @Jammjammjamm had me create the `Configuration` module to make it easier for test and development. The health_cards library also uses this `Configuration` module. Any thoughts Steve?

Jammjammjamm: The health cards library should not be using the `Configuration` module. I don't have strong feeling over using this or rails configuration, as long as the environment variable references are contained in a single place.

danlee1025: Actually now that I look at it again, the `Configuration` module is only used in the issuer tests but we can discuss how we want to do config moving forward.

ssayer (approved):

PR #20, opened 2021-03-23
Small Health Cards

mickohanlon23:
# Small Health Cards
Adds a module to constrain the size of a JWS Payload, according to:
https://smarthealth.cards/#health-cards-are-small

The `constrain_health_cards` method will take in a JWS payload, and output an encoded, compressed version of the payload after it has been stripped of extraneous resources (as defined by the spec) and minified.


lib/health_cards/health_card_constraints.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+require 'base64'
+require 'zlib'
+require 'json/minify'
+
+module HealthCards
+  ## Functionality for "Health Cards are Small" section of the Smart Health Cards Specification
+  module HealthCardConstraints
```
ssayer, line 9: Is there a reason this is done as a module? Wouldn't this be better as a part of VerifiableCredential?

lib/health_cards/health_card_constraints.rb
```diff
@@ -0,0 +1,89 @@
...
+
+        resource.delete('id')
+        resource.delete('meta')
+        resource.delete('text')
+        update_nested_elements(resource, url_map)
+      end
+      resources
+    end
+
+    def update_nested_elements(hash, url_map) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
```
ssayer, line 50: Can this be refactored or split up into methods? I'm not sure disabling rubicon here is a good idea

ssayer (changes requested):

lib/health_cards/health_card_constraints.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+require 'base64'
+require 'zlib'
+require 'json/minify'
+
+module HealthCards
+  ## Functionality for "Health Cards are Small" section of the Smart Health Cards Specification
+  module HealthCardConstraints
+    def bundle_operations(payload)
```
Jammjammjamm, line 10: This method needs a better name. Method names should usually be verbs. `bundle_operations` doesn't give enough insight into what the method does.

lib/health_cards/health_card_constraints.rb
```diff
@@ -0,0 +1,89 @@
...
+        update_elements(entries, url_map)
+      end
+      payload
+    end
+
+    def redefine_uris(resources)
+      url_map = {}
+      resource_count = 0
+
+      resources.each do |entry|
```
Jammjammjamm, line 25: If each element of this array is an `entry`, that makes me think that the array should be called `entries` rather than `resources`.

lib/health_cards/health_card_constraints.rb
```diff
@@ -0,0 +1,89 @@
...
+    def update_elements(resources, url_map)
+      resources.each do |entry|
+        resource = entry['resource']
+
+        resource.delete('id')
+        resource.delete('meta')
+        resource.delete('text')
+        update_nested_elements(resource, url_map)
+      end
+      resources
```
Jammjammjamm, line 47: You don't need to explicitly return `resources`, as `#each` returns the thing you call it on.

test/health_cards/health_card_constraints_test.rb
```diff
@@ -0,0 +1,79 @@
...
+    @dummy_class = Class.new
+    @dummy_class.extend(HealthCards::HealthCardConstraints)
+
+    file = File.read(FILEPATH_JWS_PAYLOAD)
+    jws_payload = JSON.parse(file)
+    bundle = jws_payload['vc']['credentialSubject']['fhirBundle']
+    @entries = bundle['entry']
+  end
+
+  test 'Bundle.entry.fullUrl elements are populated with short resource-scheme URIs (e.g., {"fullUrl": "resource:0})' do
```
Jammjammjamm, line 26: Rather than names like `this is done`, give your tests names like `some_method does this`.

test/health_cards/health_card_constraints_test.rb
```diff
@@ -0,0 +1,79 @@
...
+    new_entries, _url_map = @dummy_class.redefine_uris(@entries)
+    resource_nums = []
+    new_entries.each do |resource|
+      url = resource['fullUrl']
+      resource, num = url.split(':')
+      assert_equal('resource', resource)
```
Jammjammjamm, line 32: Having to assert the things you're doing in your test rather than the thing you're actually testing seems like an indication that this test is too complicated.

lib/health_cards/health_card_constraints.rb
```diff
@@ -0,0 +1,89 @@
...
+  module HealthCardConstraints
+    def bundle_operations(payload)
+      bundle = payload['vc']['credentialSubject']['fhirBundle']
+      if bundle
+        entries = bundle['entry']
+        entries, url_map = redefine_uris(entries)
+
+        update_elements(entries, url_map)
+      end
+      payload
```
Jammjammjamm, line 18: This module appears to be modifying the payload that is passed into it. I think that is very undesirable behavior.

lib/health_cards/health_card_constraints.rb
mickohanlon23: Switched name to `strip_fhir_bundle()`

lib/health_cards/health_card_constraints.rb
mickohanlon23: Changed `resources` to `entries` here and in `update_elements` method

lib/health_cards/health_card_constraints.rb
mickohanlon23: Removed.

test/health_cards/health_card_constraints_test.rb
mickohanlon23: Done.

lib/health_cards/health_card_constraints.rb
mickohanlon23: Added ticket [DVCI-53](https://jira.mitre.org/browse/DVCI-53) to update the way that FHIR Resources are navigated. According to our discussion today, I'll mark this as resolved and we can pick this back up after current PRs are merged.

test/health_cards/health_card_constraints_test.rb
mickohanlon23: Based off of all the examples I've seen in the spec, the "resource-scheme URIs" that the spec mandates each resource to have requires that each URI start with "Resource:" and then end with incremental integers. So my goal here was to check that each entry starts with "Resource", and then check that the resource numbers follow a sequential order. Should I split these two assertions into two separate tests? Or maybe only check that the _first_ resource in the bundle is equal to "Resource:0" to lesson the complexity of the test?

mickohanlon23:
Functionality from the HealthCardConstraints module has been moved to the VerifiableCredential class, so that it is more integrated with the workflow of creating a VC. Tests have been moved over to `verifiable_credential_test.rb`

ssayer (approved):

PR #21, opened 2021-03-23
DVCI-46: Add landing page and update styling

Jammjammjamm:
Screen Shot 2021-03-23 at 11 34 26 AM

Screen Shot 2021-03-23 at 11 34 35 AM

Screen Shot 2021-03-23 at 11 38 12 AM

### DIFF
--- a/app/views/immunizations/_form.html.erb
+++ b/app/views/immunizations/_form.html.erb
@@ -2,9 +2,10 @@
   <div class="container">
     <%= form_for([patient, immunization]) do |form| %>
       <% if immunization.errors.any? %>
-        <div id="error_explanation">
-          <h2><%= pluralize(immunization.errors.count, "error") %> prohibited this immunization from being saved:</h2>
-
+        <div class="notification is-danger is-light content">
+          <p class="subtitle">
+            <%= pluralize(immunization.errors.count, "error") %> prohibited this immunization from being saved:
+          </p>
           <ul>
             <% immunization.errors.each do |error| %>
               <li><%= error.full_message %></li>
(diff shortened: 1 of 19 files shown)


app/views/patients/_form.html.erb
```diff
@@ -1,40 +1,56 @@
...
+          <div class="select">
+            <%= form.select :gender, options_for_select(Patient::GENDERS)  %>
+          </div>
+        </div>
+      </div>
 
+      <div class="field">
+        <%= form.label :birth_date, class: 'label' %>
+        <div class="control">
+          <%= form.date_field :birth_date, class: 'input' %>
```
ssayer, line 42: Is there a better field we can use for this? I ran into a lot of problems with a free text field as Ruby/Rails can't always parse dates.

ssayer (changes requested):
I might have just missed it, but is there a link to the patient list anywhere? If not, there should be.

app/views/immunizations/_form.html.erb
```diff
@@ -1,32 +1,50 @@
...
+      </div>
 
-  <div class="field">
-    <%= form.label :occurrence %>
-    <%= form.date_select :occurrence %>
-  </div>
+      <div class="field">
+        <%= form.label :occurrence, class: 'label' %>
+        <div class="control">
+          <%= form.date_field :occurrence, class: 'input' %>
```
ssayer, line 28: here too (see previous comment on date_field)

app/views/patients/_form.html.erb
Jammjammjamm: You're using safari?

ssayer: Yes

Jammjammjamm: Everything other than desktop safari and IE have native date inputs. I added some more guidance and front-end validation for safari. The other option is to use a js datepicker, which I'd prefer to hold off on.

ssayer (approved):

PR #22, opened 2021-03-23
Download Health Card

ssayer:
Enables downloading of health cards.

Needs PR#20 in order to fully implement the spec


lib/health_cards/issuer.rb
```diff
@@ -1,41 +1,30 @@
...
-    def initialize(key_path = Pathname.new('.'))
-      @key_path = key_path.is_a?(String) ? Pathname.new(key_path) : key_path
+    def initialize(key_store)
+      @key = key_store.load_or_create_key
     end
 
-    def signing_key_path
-      key_path.join 'signing_key.pem'
+    def signing_key
+      @signing_key ||= JSON::JWK.new(@key, use: 'sig', alg: 'ES256')
```
danlee1025, line 17: The `json/jwt` gem defines `OpenSSL::PKey::EC#to_jwk` so I think you can remove the require and just do

```suggestion
      @signing_key ||= @key.to_jwk(use: 'sig', alg: 'ES256)
```

ssayer: OK, I'll incorporate this locally in my next push

test/health_cards/issuer_test.rb
```diff
@@ -7,26 +7,35 @@
...
+    jwks[:keys].one? do |key|
+      assert_equal 'sig', key['use']
+      assert_equal 'ES256', key['alg']
+    end
+  end
+
+  test 'created signed jwt' do
+    vc = HealthCards::VerifiableCredential.new({})
+    signed_jwt = @issuer.sign(vc, 'http://example.com')
+    assert_instance_of String, signed_jwt
```
danlee1025, line 32: Not sure if right test to put it in but can probably check that the `signed_jwt` verifies with the public key and the credential is the same.
```
    assert_equal vc.credential.as_json, JSON::JWT.decode(signed_jwt, @issuer.public_key)
```

ssayer: Good call, will add

app/models/patient.rb
```diff
@@ -23,7 +24,7 @@ def full_name
   end
 
   def to_bundle
-    entries = [json] + immunizations.map(&:json)
+    entries = ([min_json] + immunizations.map(&:min_json)).map { |e| FHIR::Bundle::Entry.new(resource: e) }
```
Jammjammjamm, line 27: Building a bundle seems like something that should be in its own class rather than being Patient's responsibility.

PR #25, opened 2021-03-31
DVCI-22: Scan QR codes

Jammjammjamm:
Adds support for scanning health cards in one or multiple QR codes. Backend support for handling these codes will be added in DVCI-30.


app/javascript/packs/qr_code_scan.js
```diff
@@ -0,0 +1,132 @@
...
+  errorNotification.setAttribute('hidden', true);
+};
+
+const showErrorNotification = () => {
+  hideSuccessNotification();
+  errorNotification.removeAttribute('hidden');
+};
+
+const handleScan = result => {
+  console.log(result);
```
radamson, line 74: Is this intentional or left over from debugging?

app/javascript/packs/qr_code_scan.js
```diff
@@ -0,0 +1,132 @@
...
+    const match = result.match(healthCardPattern);
+    if(match.groups.multipleChunks) {
+      hideErrorNotification();
+      const chunkCount = +match.groups.chunkCount;
+      const currentChunkIndex = +match.groups.chunkIndex;
+      if(scannedCodes.length !== chunkCount) {
+        scannedCodes = new Array(chunkCount);
+        for(let i = 0; i < chunkCount; i++) {
+          scannedCodes[i] = null;
+        }
```
radamson, line 86: Why does this Array need to be populated with `null`, instead of the leaving the entries as `undefined` (which is also falsy)?

If this is needed, could we use [`fill()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill) instead?

app/javascript/packs/qr_code_scan.js
```diff
@@ -0,0 +1,132 @@
...
+      }
+      scannedCodes[currentChunkIndex - 1] = result;
+      multiStatusContainer.innerHTML =
+        scannedCodes
+        .map((code, index) => {
+          return code ?
+            multiPresentElement(index +1, chunkCount) : multiMissingElement(index + 1, chunkCount);
```
radamson, line 93:
```suggestion
            multiPresentElement(index + 1, chunkCount) : multiMissingElement(index + 1, chunkCount);
```

app/javascript/packs/qr_code_scan.js
Jammjammjamm: Intentional. I can remove it, I just thought it was helpful for now to be able to see the contents in the browser.

app/javascript/packs/qr_code_scan.js
Jammjammjamm: Yeah, we can use `fill`. Even though the array is created with a specific length, and returns that length if you ask for it, it won't iterate over those elements without them being initialized.

Screen Shot 2021-04-01 at 2 40 26 PM

danlee1025:
Something that I've noticed is that clicking the Start button does not immediately work when I go to the "Scan a QR Code" page from the landing page. I usually have to refresh the page before I can start scanning. Have you noticed this behavior?

Jammjammjamm:
I pushed up a fix. Good old turbolinks.

mickohanlon23 (approved):
Looked through the code changes and everything looks good to me! Successful scan, multiple QR code scan, and 'Not a Smart Health Card' scan all seem to work as expected.

danlee1025 (approved):

PR #28, opened 2021-04-07
Display contents after scanning QR codes

danlee1025:
- Added logic to assemble QR code chunks and get jws payload.
- Added page to display Patient and Immunization data
    - User is redirected to this page after clicking the "View Contents" button in the scan qr code page.
    - Page looks very similar to Patient show page without the edit and destroy buttons.

This is currently branched off the scan-qr-codes PR and I used examples from https://smarthealth.cards/examples/ to test.


app/controllers/health_cards_controller.rb
```diff
@@ -19,12 +19,16 @@ def scan; end
...
 
-    Rails.logger.info(contents.join("\n"))
+    patient_resource = bundle.entry.select { |e| e.resource.is_a?(FHIR::Patient) }[0].resource
+    @pat = Patient.new(json: patient_resource)
+    create_immunizations bundle
 
-    # TODO: decode the contents and display them
-    flash[:notice] = 'Viewing contents not yet implemented'
-    redirect_to scan_health_card_path
+    respond_to do |format|
```
Jammjammjamm, line 29: This shouldn't be necessary.

app/controllers/health_cards_controller.rb
```diff
@@ -19,12 +19,16 @@ def scan; end
 
   def qr_contents
     contents = JSON.parse(params[:qr_contents])
+    jws_payload = HealthCards::Chunking.get_payload_from_qr contents
+    bundle = FHIR.from_contents(jws_payload['vc']['credentialSubject']['fhirBundle'].to_json)
```
Jammjammjamm, line 23: The code to process the qr contents shouldn't live in the controller. Extracting the contents of a QR code is a piece of business logic that should be executable and testable outside the context of an http request/response.

app/controllers/health_cards_controller.rb
```diff
@@ -19,12 +19,16 @@ def scan; end
...
 
-    Rails.logger.info(contents.join("\n"))
+    patient_resource = bundle.entry.select { |e| e.resource.is_a?(FHIR::Patient) }[0].resource
```
Jammjammjamm, line 25: I get this error when using [example 2](https://smarthealth.cards/examples/#example-2):

Screen Shot 2021-04-09 at 9 59 32 AM

danlee1025: Yea I'm expecting a `Patient` resource in the fhir bundle. [Example 2](https://smarthealth.cards/examples/example-02-a-fhirBundle.json) doesn't seem to have one. Do you think I should just display the whole vc if no patient is available?

Jammjammjamm: Sure.

mickohanlon23: I think this makes the most sense. Fill in the patient and immunization into the nice UI if they are available, but default to displaying the VC if they are not available since they aren't actually required by SMART Health Cards.

lib/health_cards/chunking.rb
```diff
@@ -23,11 +23,47 @@ def generate_qr_chunks(jws)
...
+        # Strip off shc:/ and convert numeric jws
+        numeric_jws = qr_chunks[0].delete_prefix('shc:/')
+        convert_numeric_jws numeric_jws
+      else
+        ordered_qr_chunks = strip_prefix_and_sort qr_chunks
+        ordered_qr_chunks.map { |c| convert_numeric_jws(c) }.join
+      end
+    end
+
+    def get_payload_from_qr(qr_chunks)
```
mickohanlon23, line 38: This may depend on some of the refactor code currently in progress, but should the issuer and verifier methods in chunking.rb be split into their own classes instead of living in the same module?

app/controllers/health_cards_controller.rb
```diff
@@ -43,4 +47,22 @@ def health_card
...
+
+  def create_immunizations(bundle)
+    immunizations = bundle.entry.select { |e| e.resource.is_a?(FHIR::Immunization) }
+    vaccine_id_map = {
+      '207': 1,
+      '208': 2,
+      '212': 3
+    }
+
+    immunizations.each do |i|
```
mickohanlon23, line 59: Will this still work completely fine if there is no immunization resource?

app/controllers/health_cards_controller.rb
```diff
@@ -43,4 +47,22 @@ def health_card
...
+      immunization_resource = i.resource
+      @pat.immunizations.new({
+                               vaccine_id: vaccine_id_map[immunization_resource.vaccineCode.coding[0].code.to_sym],
```
mickohanlon23, line 62: Similarly will there be errors if these elements aren't present? I know they are required by the FHIR IG, but for general Smart Health Cards we might want to only initialize an immunization if those data minimization elements are present in the bundle that is passed in.

mickohanlon23 (commented):
Looks great. I reviewed against the Chunking section of the spec and the logic all makes sense. I left a few inline comments, but other than that, this looks good!

app/controllers/health_cards_controller.rb
Jammjammjamm: I think we are specifically doing a Covid-19 implementation, so if `Patient` is required, it's ok to display a message explaining that it isn't a valid VC, especially as a first pass. It's just not ok to return a 500.

app/controllers/health_cards_controller.rb
danlee1025: Moved this logic to helper

app/controllers/health_cards_controller.rb
danlee1025: Changed to display jws payload if `Patient` isn't found in bundle.

app/controllers/health_cards_controller.rb
danlee1025: It should just be an empty array if there are no immunization resources so it should be fine. The code inside the `each` just won't be executed.

app/controllers/health_cards_controller.rb
danlee1025: I think the `vaccineCode` and `occurenceDateTime` have to be in the immunization. I think it will just an empty string if `lotNumber` is not present.

danlee1025:
@Jammjammjamm Updated jws_payload view to display an HTML page instead of just rendering json when `Patient` wasn't in bundle.  Did we want to use something like [json-viewer-js](https://www.npmjs.com/package/json-viewer-js) to display the json and make it easier to navigate?

Jammjammjamm:
I think it's fine as-is for now. We can add that later if necessary.

Jammjammjamm (approved):

lib/health_cards/chunking.rb
danlee1025: Yea I think this will probably depend on the refactor. I think for now the code is fine to live in the `Chunking` module.

mickohanlon23: I agree.

mickohanlon23 (approved):
Updated view-page for a Bundle without a patient looks great! Good job!

PR #29, opened 2021-04-13
Meta special case

mickohanlon23:
## Small change in the way bundle stripping is done, according to https://github.com/smart-on-fhir/health-cards/pull/117/files:

Resource.meta fields are now allowed to be included within FHIR bundles if they contain only a `security` element. 

This PR also fixes a bug in one of the VerifiableCredential tests where the test was not actually checking to see if the resource-level elements were removed (instead of checking if `Resource.text`, `Resource.meta`, and `Resource.id` were removed, it was doing this check one-level up, so it was only actually checking the `resource` and `fullUrl` fields of each Bundle entry.

### DIFF
--- a/lib/health_cards/verifiable_credential.rb
+++ b/lib/health_cards/verifiable_credential.rb
@@ -89,7 +89,11 @@ module HealthCards
         resource = entry['resource']
         resource.delete('id')
         resource.delete('text')
-        resource['meta'] = resource['meta'].slice('security') if resource.key?('meta')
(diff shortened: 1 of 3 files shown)


mickohanlon23:
Made a small update according to a conversation yesterday with @Jammjammjamm. https://github.com/smart-on-fhir/health-cards/pull/117 was merged today, so this should be good for review.

Jammjammjamm (approved):

PR #30, opened 2021-04-27
Custom Health Cards

ssayer:
This PR
* Separates the COVID IG specific implementation out into its own class that subclasses HealthCard. 
* Merges VerifiableCredential with HealthCard
* Removes the JWS instance variable from HealthCard


app/controllers/health_cards_controller.rb
```diff
@@ -25,18 +25,30 @@ def qr_contents
 
   private
 
+  def jws
```
Jammjammjamm, line 28: All of this business logic should live somewhere outside of the controller.

lib/health_cards/covid_health_card.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module HealthCards
+  # Implements HealthCard for use with COVID Vaccination IG
+  class COVIDHealthCard < HealthCards::HealthCard
+    fhir_version '4.0.1'
+    additional_types 'https://healthwallet.cards#covid19'
+
+    allow FHIR::Patient, %w[name birthDate]
+    allow FHIR::Immunization, %w[status vaccineCode patient occurrenceDateTime]
```
Jammjammjamm, line 10: This looks really nice, but I worry about how complicated this might have to become.

lib/health_cards/health_card.rb
```diff
@@ -11,43 +17,87 @@ class << self
...
+        bundle_hash = json.dig('vc', 'credentialSubject', 'fhirBundle')
+
+        raise InvalidCredentialError unless bundle_hash
+
+        bundle = FHIR::Bundle.new(bundle_hash)
+        new(issuer: json['iss'], bundle: bundle)
+      end
+
+      def allow(klass, attributes)
+        @allowable ||= {}
```
Jammjammjamm, line 50: You don't need to do this since you have the method below.

lib/health_cards/health_card.rb
```diff
@@ -11,43 +17,87 @@ class << self
...
+
+      def fhir_version(ver = nil)
+        @fhir_version ||= ver unless ver.nil?
+        @fhir_version
       end
-    end
 
-    attr_reader :verifiable_credential, :jws
+      def additional_types(*types)
+        @types ||= VC_TYPE
```
Jammjammjamm, line 75: Again, you can just use the method you created here.

lib/health_cards/health_card.rb
```diff
@@ -1,64 +1,208 @@
...
-    # @param file_name [String] the name of the file
-    def save_to_file(file_name)
-      File.open(file_name, 'w') do |file|
-        file.write(to_json)
-      end
+    # A compressed version of the FHIR::Bundle based on the SMART Health Cards frame work and any other constraints
+    # defined by a subclass
+    # @return String compressed payload
+    def to_s
+      HealthCard.compress_payload(to_json)
```
Jammjammjamm, line 118: Why is `compress_payload` a class method?

lib/health_cards/health_card.rb
```diff
@@ -1,64 +1,208 @@
...
+    end
+
+    # Chunks the compress payload for us in generating QR codes
+    # @return [Array] An array of strings used to create QR codes
+    def chunks
+      HealthCards::Chunking.generate_qr_chunks to_s
```
Jammjammjamm, line 124: Should we just `include` this module?

Jammjammjamm (commented):
I think this looks like a nice improvement.

lib/health_cards/covid_health_card.rb
ssayer: Yeah, I spoke to @radamson about this. We agreed that we'll deal with that when we need to. I think any crazy requirements we run into in the future would require an equally crazy implementation anyways.

app/controllers/health_cards_controller.rb
ssayer: I agree with that and it pained me to put this in here, but we need access to the url helpers to create the fullUrls for the bundle and I can't think of a better place to do it. I used to do that in CovidHealthCard because it lived in Railsland, but now it's in the library. I didn't want to force the library to do a bunch of weird url manipulation because of rails

Jammjammjamm: I'm fine with it living in rails, I just don't think it should live in a controller.

lib/health_cards/health_card.rb
ssayer: Maybe? I'm not even sure why it's a module as it's only used in health card

lib/health_cards/health_card.rb
ssayer: HealthCard handles all the payload manipulation and for testing purposes I needed to be able to compress arbitrary payloads.

lib/health_cards/health_card.rb
ssayer: If we do want to change how we do this though, I'd consider that out of scope for this task.

lib/health_cards/health_card.rb
ssayer: will fix

lib/health_cards/health_card.rb
```diff
@@ -47,7 +47,7 @@ def compress_payload(payload)
       # @param attributes [Array] An array of string with the attribute names that will be passed through
       #  when data is minimized
       def allow(klass, attributes)
-        @allowable ||= {}
+        allowable
```
Jammjammjamm, line 50: You don't need this line, you just need to use `allowable` below instead of `@allowable`. Same thing with `types` in `#additional_types`.

ssayer: Isn't that what it's doing now?

Jammjammjamm:
```ruby
def allow(klass, attributes)
  resource_type = klass.name.split('::').last
  allowable[resource_type] = attributes
end

def additional_types(*add_types)
  types += add_types
end
```

app/controllers/health_cards_controller.rb
ssayer: OK, now it is done in patient

lib/health_cards/health_card.rb
Jammjammjamm: This hasn't been addressed.

ssayer: Oops, I didn't see your follow up comment. This should be fixed now.

Jammjammjamm (approved):

PR #31, opened 2021-04-29
Better reference handling in Bundles

ssayer:
Key Changes/Features
* Assigns ids to the FHIR JSON in our database which weren't there before. This id matches the database id. There is a migration that will need to be run for existing records to insert the id.
* Our reference processing will now handle logical references (e.g. Patient/1 -> http://example.org/Patient/1) provided urls are relative to the issuer url contained in the HealthCard
* If our bundle processing can't resolve a reference within a bundle it will now throw an exception as it is expected that all references resolve within the bundle.
* Cleaned up the Bundle tree walking code to remove Rubocop Perceived Complexity warning. Cyclomatic Complexity is lowered but still higher than default Rubocop setting.

### DIFF
--- a/app/models/fhir_record.rb
+++ b/app/models/fhir_record.rb
@@ -6,6 +6,8 @@ class FHIRRecord < ApplicationRecord
 
   validate :valid_fhir_json
 
+  after_create :set_fhir_id
+
   def to_json(*_args)
     json.to_hash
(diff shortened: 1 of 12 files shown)


ssayer:
Reviewed with @Jammjammjamm and @radamson offline

PR #32, opened 2021-04-30
Update Migration

radamson:
Right now the `AddIdToFHIRJson` is broken because it tries to use the protected `set_fhir_id` method. This PR fixes that issue by defining new migration specific models to use.

See [Rails Style Guide Define Model Class](https://rails.rubystyle.guide/#define-model-class-migrations) and [Ruby on Rails Using Models in Migrations](https://guides.rubyonrails.org/v4.1/migrations.html#using-models-in-your-migrations)

Side Note: Interestingly the [new Rails Guide doesn't have a section on using models in migrations](https://guides.rubyonrails.org/active_record_migrations.html). Is it better to use straight up SQL here?

### DIFF
--- a/db/migrate/20210428145300_add_id_to_fhir_json.rb
+++ b/db/migrate/20210428145300_add_id_to_fhir_json.rb
@@ -1,28 +1,38 @@
 # frozen_string_literal: true
 
+
 # Sets ID in existing FHIR-based records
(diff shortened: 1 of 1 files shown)


ssayer (approved):
Reverted mine and reran and it seems to work

Jammjammjamm (approved):

PR #33, opened 2021-04-30
Fix QR code display

danlee1025:
- Changed to generate qr chunks from JWS string instead of health card compressed payload.

Currently branched off of `update_migration` PR to get the app running.

### DIFF
--- a/app/controllers/health_cards_controller.rb
+++ b/app/controllers/health_cards_controller.rb
@@ -12,7 +12,7 @@ class HealthCardsController < ApplicationController
   end
 
   def chunks
-    render json: HealthCards::Chunking.generate_qr_chunks(jws.to_s)
+    render json: jws.chunks
   end
 
   def scan; end
--- a/app/controllers/health_cards_controller.rb
+++ b/app/controllers/health_cards_controller.rb
@@ -12,7 +12,7 @@ class HealthCardsController < ApplicationController
   end
 
   def chunks
-    render json: health_card.chunks.to_json
+    render json: HealthCards::Chunking.generate_qr_chunks(jws.to_s)
   end
 
   def scan; end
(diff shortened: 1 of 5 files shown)


ssayer (changes requested):
This looks good to me, but can we add a test so in the future if something changes we'll know?

app/controllers/health_cards_controller.rb
```diff
@@ -12,7 +12,7 @@ def show
   end
 
   def chunks
-    render json: health_card.chunks.to_json
+    render json: HealthCards::Chunking.generate_qr_chunks(jws.to_s)
```
ssayer, line 15: I think it would make sense to add this code to jws.rb instead of calling the module directly.

test/controllers/health_cards_controller_test.rb
```diff
@@ -45,7 +45,10 @@ class HealthCardsControllerTest < ActionDispatch::IntegrationTest
     get(chunks_patient_health_card_url(@patient))
     assert_response :success
 
-    json = JSON.parse(response.body)
-    assert_equal 1, json.size
+    chunks = JSON.parse(response.body)
+    # Check that each string in the array is a numeric string
+    chunks.each do |s|
+      assert_empty(s.scan(/\D/))
```
ssayer, line 51: Can we do any other assertions based on the contents aside from numeric?

app/controllers/health_cards_controller.rb
danlee1025: Moved it to `jws.rb`

test/controllers/health_cards_controller_test.rb
danlee1025: Yep, I've updated the test to reconstruct the jws and check that the bundle entries match.

ssayer (approved):

PR #36, opened 2021-05-04
issue health card endpoint

ssayer:
Enabled $issue endpoint using refactored code


lib/health_cards/health_card.rb
```diff
@@ -69,16 +69,26 @@ def fhir_version(ver = nil)
...
 
       # Type claims supported by this HealthCard subclass
       # @return [Array] an array of Strings with all the supported type claims
       def types
         @types ||= VC_TYPE.dup
       end
+
+      # Check if this class supports the given type
+      # @param type [String] A type as defined by the SMART Health Cards framework
+      # @return [Array] All additional types supported by this HealthCard (sub)class
```
Jammjammjamm, line 88: This needs updating.

app/controllers/health_cards_controller.rb
```diff
@@ -7,7 +7,34 @@ class HealthCardsController < ApplicationController
 
   def show
     respond_to do |format|
-      format.healthcard { render json: { verifiableCredential: [jws.to_s] } }
+      vc = jws.to_s
+      format.healthcard { render json: { verifiableCredential: [vc] } }
+      format.fhir_json do
+        fhir_params = FHIR.from_contents(request.raw_post)
```
Jammjammjamm, line 13: For this code, I'd suggest we add an `app/lib` directory, and stick classes in there to handle this. So this becomes something along the lines of:
```
format.healthcard { render json: HealthCardsFileIssuer.new(some_params).issue }
format.fhir_json { render json: HealthCardsFHIRIssuer.new(some_other_params).issue }
```

lib/health_cards/health_card.rb
ssayer: Do you mean the documentation or something else? Just want to make sure I'm not missing something

app/controllers/health_cards_controller.rb
ssayer: I think we can move a lot of the rails specific code to something in `app/lib` but I think a good portion of this functionality is non-rails specific and could benefit implementers if it's in the library. @radamson - Thoughts?

lib/health_cards/health_card.rb
Jammjammjamm: Sorry, just the documentation.

app/controllers/health_cards_controller.rb
radamson: To recap our discussion with @Jammjammjamm I think that.

- The FHIR Request Parameter checks should stay out of the `health_cards` library, but it could be moved into `lib` so that logic can stay out of the controller.
- Create an `Exporter` module in the `health_cards` library that takes in a jws and returns in in the desired export format (e.g. in a parameters resource, wrapped in json for a file download)

ssayer:
I still need to do a little cleanup and documentation, but wanted to make sure this matches expectations from our discussion

lib/health_cards/health_card.rb
```diff
@@ -72,22 +72,32 @@ def fhir_version(ver = nil)
...
 
       # Type claims supported by this HealthCard subclass
       # @return [Array] an array of Strings with all the supported type claims
       def types
         @types ||= VC_TYPE.dup
       end
+
+      # Check if this class supports the given type claim(s)
+      # @param type [Array, String] A type as defined by the SMART Health Cards framework
+      # @return [Bool] Whether or not the type param is included in the types supported by the HealthCardn (sub)class
```
Jammjammjamm, line 91:
```suggestion
      # @return [Boolean] Whether or not the type param is included in the types supported by the HealthCard (sub)class
```

Jammjammjamm (approved):

PR #37, opened 2021-05-06
Upload Credentials

ssayer:
Add form on main page to upload smart-health-card files and render results.

NOTE: I had to make some minor changes to QR scanning code in order to reuse it for file upload, so please review to ensure things are working as expected there. My webcam doesn't seem to work with the web api (tried before I made changes).


Jammjammjamm (commented):
I get this error using the first two examples from the health card spec, when uploading or using a QR code. The third example works fine.

Screen Shot 2021-05-06 at 1 09 49 PM

danlee1025:
Examples worked for me when scanning QR code and uploading. @Jammjammjamm Does your DB need an update? Error looks like it's having trouble finding the vaccine.

Jammjammjamm:
Resetting my db did fix this. I guess this was a consequence of that other branch where we had to change a migration.

Jammjammjamm (approved):

PR #39, opened 2021-05-07
Key Resolution

radamson:
# Public Key Resolution

This PR implements public key resolution for Health Cards by enabling the `HealthCards::Verifier` to query the issuer's `.well-known/jwks.json` endpoint. Key resolution is enabled by default for all `HealthCards::Verifier` instances and for `HealthCards::Verifier::verify` class method.

@ssayer https://github.com/dvci/health_cards/commit/915536e55e177abc410b34f51e1882f73964f682 changes the `HealthCards::Key#coordinates` method to work directly with the binary public key representation instead of the hex version (which then needed to be `pack('H*')`ed)

**New Methods**

- `HealthCards::Key::from_jwk`
- `HealthCards::KeySet::from_jwks`
- `HealthCards::Verifier::verify`
- `HealthCards::Verifier::resolve_key`
- `HealthCards::Verifier#resolve_keys?`
  - Also includes accessors `HealthCards::Verifier#resolve_keys` and `HealthCards::Verifier#resolve_keys=`

## References
[SMART Health Cards Framework: Determining keys associated with an issuer](https://smarthealth.cards/#determining-keys-associated-with-an-issuer)


lib/health_cards/jws.rb
```diff
@@ -19,11 +19,17 @@ def decode(data)
...
       # @param public_key [HealthCards::PublicKey] the public key associated with the JWS
       # @param key [HealthCards::PrivateKey] the private key associated with the JWS
       # @return [HealthCards::HealthCard]
       def from_jws(jws, public_key: nil, key: nil)
+        unless jws.is_a?(HealthCards::JWS) || jws.is_a?(String)
+          raise ArgumentError,
+                'Expected either a HealthCards::JWS or String'
+        end
+
+        jws = jws.to_s
```
ssayer, line 32: This modifies the parameter. Recommended putting it below when you call `split`

lib/health_cards/key.rb
```diff
@@ -16,6 +16,20 @@ def self.enforce_valid_key_type!(obj, allow_nil:  false)
       raise InvalidKeyException.new(self, obj) unless obj.is_a?(self) || (allow_nil && obj.nil?)
     end
 
+    # Create a key from a JWK
+    #
+    # @param jwk_key [Hash] The JWK represented by a Hash
+    # @return [HealthCards::Key] The key represented by the JWK
+    def self.from_jwk(jwk_key)
+      jwk_key = jwk_key.transform_keys(&:to_sym)
+      group = OpenSSL::PKey::EC::Group.new('prime256v1')
```
ssayer, line 25: Since we're always using this group, maybe make this a constant.

Jammjammjamm: I don't think that is safe unless `Group` is immutable.

lib/health_cards/verifier.rb
```diff
@@ -4,11 +4,35 @@ module HealthCards
...
+    #
+    # This method _always_ uses key resolution and does not depend on any cached keys
+    #
+    # @param verifiable [HealthCards::JWS, String] the health card to verify
+    # @return [Boolean]
+    def self.verify(verifiable)
+      jws = JWS.from_jws(verifiable)
+      key_set = resolve_key(jws)
+      key = key_set.find_key(jws.kid)
+      jws.public_key = key
```
ssayer, line 19: This code seems at least somewhat duplicative of the instance method

radamson: I created a `verify_using_key_set` method to hold the shared logic.

First (This commit: https://github.com/dvci/health_cards/pull/39/commits/7c478f65f9a34879a7ed1df46d196858a6058fbd)  I just added this as a class method that the instance `verify` method could also call

Then (This commit: https://github.com/dvci/health_cards/pull/39/commits/1b1b4722dd00df3fed7a520fc021b59da15280ee) I moved `verify_using_key_set` and `resolve_key` to a new module so that I could just `include`/`extend` `Verifier` with it (seemed a little cleaner)

Do you prefer one approach over another or have a different suggestion?

Jammjammjamm (approved):

ssayer (approved):

PR #42, opened 2021-05-11
Details Page for Each Patient

pmadria:
Details page shows for each patient 4 things: 

- Fhir Bundle
- Minified Fhir Bundle
- Encoded JWS 
- QR String of their numerical array for the JWS 

The show page then has an expand/collapse feature showing the details.


app/javascript/packs/details_pg.js
```diff
@@ -0,0 +1,13 @@
+function onReady() {
+    let cardToggles = document.getElementsByClassName('card-toggle');
+	for (let i = 0; i < cardToggles.length; i++) {
+		cardToggles[i].addEventListener('click', e => {
+			e.currentTarget.parentElement.parentElement.childNodes[3].classList.toggle('is-hidden');
```
Jammjammjamm, line 5: Something weird is going on with the indentation in this file.

lib/health_cards/jws.rb
```diff
@@ -24,7 +24,7 @@ def decode(data)
       # @param key [HealthCards::PrivateKey] the private key associated with the JWS
       # @return [HealthCards::HealthCard]
       def from_jws(jws, public_key: nil, key: nil)
-        header, payload, signature = jws.split('.').map { |entry| decode(entry) }
+        header, payload, signature = jws.to_s.split('.').map { |entry| decode(entry) }
```
Jammjammjamm, line 27: The documentation for this method says `jws` should be a string, so we shouldn't have to call `to_s` on it.

app/views/health_cards/show.html.erb
```diff
@@ -0,0 +1,82 @@
...
+    <div class="card is-fullwidth">
+      <header class="card-header">
+        <p class="card-header-title">Minified FHIR Bundle</p>
+        <a class="card-header-icon card-toggle">
+          <i class="fa fa-angle-down"></i>
+        </a>
+      </header>
+        <div class="card-content">
+          <div class="content">
+           <pre>
```
Jammjammjamm, line 38: I think all of these should be wrapped in `pre` blocks.

app/views/health_cards/show.html.erb
```diff
@@ -0,0 +1,82 @@
...
+              <%= @health_card.bundle.to_json %>
+           </pre>
+          </div>
+        </div>
+      </div>
+    <br>
+    <br>
+    <br>
+
+    <p class="title">Encoded JWS</p>
```
Jammjammjamm, line 48: The titles don't need to be repeated here and as card headers.

app/views/health_cards/show.html.erb
```diff
@@ -0,0 +1,82 @@
...
+    <div class="card is-fullwidth">
+      <header class="card-header">
+        <p class="card-header-title">QR Details</p>
+        <a class="card-header-icon card-toggle">
+          <i class="fa fa-angle-down"></i>
+        </a>
+      </header>
+        <div class="card-content">
+          <div class="content", style="word-wrap: break-word; width: 800px">
+            <%= @qr_code_payload %>
```
Jammjammjamm, line 75: I think the entire qr code payload, including the prefix, should be displayed here. Also, they shouldn't be displayed as raw arrays.

Jammjammjamm (commented):
There are rubocop errors that need to be fixed.

Screen Shot 2021-05-12 at 9 30 51 AM

Are you sure that the bundle and minified bundle are in the right places? It seems odd to me to display the minified bundle with indentation, but to display the non-minified bundle without indentation. Also, the first line of the minified bundle has weird indentation.

Screen Shot 2021-05-12 at 9 38 25 AM

This button shouldn't have zero margin.

app/controllers/health_cards_controller.rb
```diff
@@ -20,6 +25,11 @@ def create
         fhir_params = FHIR.from_contents(request.raw_post)
         render json: @exporter.issue(fhir_params)
       end
+      format.html do
+        @jws_encoded_details = @exporter.jws
+        @health_card = HealthCards::HealthCard.from_jws @jws_encoded_details.to_s
```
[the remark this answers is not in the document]
pmadria: yeah I had a rebase nightmare and it I guess put them in both places...fixed it and moved it to show

lib/health_cards/jws.rb
ssayer: This method should probably also accept a JWS object since we often have to chain calls to get the JWS and then the HealthCard objects, so recommend updating the documentation

app/controllers/health_cards_controller.rb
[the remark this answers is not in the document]
radamson: @pmadria is this actually resolved? It looks like `.to_s` is still on this line, but not line 16. Maybe my diff is out of date, but is there a reason for this?

app/controllers/health_cards_controller.rb
```diff
@@ -33,7 +43,6 @@ def qr_contents
     @jws_payload = HealthCards::Importer.scan(params[:qr_contents])
     @patient = helpers.create_patient_from_jws(@jws_payload)
   end
-
```
radamson, line 36: Rubocop is complaining about this missing empty line:

```
app/controllers/health_cards_controller.rb:46:3: C: [Correctable] Layout/EmptyLineBetweenDefs: Expected 1 empty line between method definitions; found 0.
  def upload
  ^^^^^^^^^^

```

app/views/health_cards/show.html.erb
radamson: @pmadria is this actually resolved? It looks like the title is still repeated

Screen Shot 2021-05-14 at 11 13 19 AM

app/views/health_cards/show.html.erb
pmadria: resolved

app/controllers/health_cards_controller.rb
pmadria: resolved

app/views/health_cards/show.html.erb
pmadria: resolved

lib/health_cards/jws.rb
pmadria: resolved

app/javascript/packs/details_pg.js
pmadria: resolved

app/controllers/health_cards_controller.rb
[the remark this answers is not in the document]
pmadria: resolved

app/javascript/packs/details_pg.js
```diff
@@ -0,0 +1,14 @@
+if (document.readyState !== "loading") {
+    onReady();
+} else {
+    document.addEventListener("DOMContentLoaded", onReady);
+}
+
+function onReady() {
```
mickohanlon23, line 7: It looks like the indentation in this file is still 4 spaces when it should be 2.

app/views/health_cards/show.html.erb
```diff
@@ -0,0 +1,83 @@
+<%= javascript_pack_tag 'details_pg' %>
+<p id="notice"><%= notice %></p>
+<section class="hero is-link">
+  <div class="hero-body">
+    <p class="title">Patient Details</p>
```
mickohanlon23, line 5: The button to get to this page is called "Health Card Details", but then this page is titled "Patient Details". We should make these consistent. I think it makes sense to title this page "Health Card Details", Since it is showing how the actual health card was built up, not the patient resource (the patient show page where the button lives would technically be the "Patient Details" page).

app/views/health_cards/show.html.erb
```diff
@@ -0,0 +1,83 @@
...
+      <header class="card-header">
+        <p class="card-header-title">Minified FHIR Bundle</p>
+        <a class="card-header-icon card-toggle">
+          <i class="fa fa-angle-down"></i>
+        </a>
+      </header>
+        <div class="card-content">
+          <div class="content">
+           <pre>
+              <p> <%= @health_card.bundle.to_json %> </p>
```
mickohanlon23, line 40: Minifying a JSON file means that all whitespace is removed (indents, newlines, spaces, etc.). It looks like what you're displaying here is the FHIR Bundle before it is actually minified.

app/views/health_cards/show.html.erb
```diff
@@ -0,0 +1,83 @@
...
+            </pre>
+            </div>
+          </div>
+        </div>
+    <br>
+    <br>
+
+    <div class="card is-fullwidth">
+      <header class="card-header">
+        <p class="card-header-title">Minified FHIR Bundle</p>
```
mickohanlon23, line 32: Technically the FHIR Bundle isn't minified, but the JWS Payload is minified, so I think we'll want to change the title here. Also the displayed minified object here should include all the top-level JWS fields, not just the FHIR Bundle.

mickohanlon23: https://smarthealth.cards/examples/example-00-c-jws-payload-minified.json

Here is a good example from the spec of what this should look like.

app/views/health_cards/show.html.erb
```diff
@@ -0,0 +1,83 @@
...
+          <div class="content">
+            <pre style="white-space:pre-wrap;">
+              <p> <%= JSON.pretty_generate(@health_card.to_hash) %> </p>
+            </pre>
+            </div>
+          </div>
+        </div>
+    <br>
+    <br>
+
```
mickohanlon23, line 29: I think it would be helpful to add one more dropdown here for the JWS Payload before it is minified (this would correspond with [example b](https://smarthealth.cards/examples/example-00-b-jws-payload-expanded.json) from the spec). Technically you already have the data for that right now displaying under the "FHIR Bundle" toggle.

app/controllers/health_cards_controller.rb
```diff
@@ -11,6 +11,19 @@ class HealthCardsController < ApplicationController
   def show
     respond_to do |format|
       format.healthcard { render json: @exporter.download }
+      format.html do
+        @jws_encoded_details = @exporter.jws
+        @health_card = HealthCards::COVIDHealthCard.from_jws @jws_encoded_details
+        @qr_code_payload = @exporter.chunks
+
+        if @qr_code_payload.length == 1
+          @qr_code_payload[0] = "shc:/#{@qr_code_payload[0]}"
```
mickohanlon23, line 20: I believe we should remove the quotes and brackets when we display this. Have the display start with `shc:/` instead of `["shc:/`. When I tested this with the Microsoft Verifier portal, I had to delete both of those to get it to decode. This also aligns with [example f](https://smarthealth.cards/examples/example-00-f-qr-code-numeric-value-0.txt) from the spec.

app/views/health_cards/show.html.erb
```diff
@@ -0,0 +1,83 @@
...
+            </div>
+          </pre>
+        </div>
+      </div>
+    <br>
+    <br>
+    
+    <div class="card is-fullwidth">
+      <header class="card-header">
+        <p class="card-header-title">QR Details</p>
```
mickohanlon23, line 68: I think this should be called something like "Numeric QR", since it's not actually showing much detail.

mickohanlon23 (changes requested):
Overall, I think it makes the most sense to align these SMART health cards building "steps" to how they are displayed in the [spec examples](https://smarthealth.cards/examples/). It looks like right now, you mostly have everything we would want to show (a, b, c, and f), I would just add one drop-down to show the actual expanded JWS (d). Some of the titles don't exactly correspond with what is being showed, so most of my comments have to do with the actual contents that we are displaying on the page.

pmadria:
Updated with changes for the latest comments

app/views/health_cards/show.html.erb
```diff
@@ -0,0 +1,103 @@
...
+            <pre style="white-space:pre-wrap;">
+              <p> <%= @health_card.to_json %> </p>
+            </pre>
+            </div>
+          </div>
+        </div>
+    <br>
+    <br>
+
+    <div class="card is-fullwidth">
```
mickohanlon23, line 48: Could we move the FHIR Bundle as the top thing on this page, since this is what is created first? Other than that, it looks like everything else is in order.

pmadria: ah good point I renamed and forgot to move it...moved to the top

mickohanlon23 (approved):
Looks great!

PR #43, opened 2021-05-13
Download PDF

mickohanlon23:
### Adds button to the patient show page allowing the user to download a PDF version of their Smart Health Card.


mickohanlon23:
Currently running into two issues I'm hoping people can take a look at:
1. When I actually try to download the PDF to my disk in Chrome, I receive a "Failed - Network error" message. However, I am able to properly download when using Safari. Are people having the same issue on their machines?
2. This should be an easy fix, but the "Download PDF" button isn't spaced properly along with the other two buttons on the patient show page.

radamson:
> 1. I actually try to download the PDF to my disk in Chrome, I receive a "Failed - Network error" message. However, I am able to properly download when using Safari. Are people having the same issue on their machines?

I see this too. It looks like when you click the download down arrow that chrome is actually trying to access the resource directly from the server

Screen Shot 2021-05-14 at 1 04 29 PM

using the Chrome DevTools I see `Failed to load resource: net::ERR_FILE_NOT_FOUND`. Im guessing chrome is actually trying to access this file but its not available from a `GET http://server.com/patients/1/health_card.pdf`, only from a `POST http://server.com/patients/1/health_card.pdf` with form data that includes the qr code. You can work around the download issue by doing a Print > Save as PDF, but it's not a real replace meant. (And I'm not exactly sure this is what is going on since when I look in the network tab it doesn't actually seem like a subsequent request is made...)

app/views/patients/show.html.erb
```diff
@@ -11,6 +11,10 @@
           <div class="is-grouped">
             <%= link_to 'Edit', edit_patient_path(@patient), class: 'button is-info is-light' %> 
             <%= link_to 'Download Health Card', patient_health_card_path(@patient, format: 'smart-health-card'), class: 'button is-info is-light' %>
+            <%= form_with url: patient_health_card_path(@patient, format: 'pdf'), method: :post do |f| %>
```
radamson, line 14:
```suggestion
            <%= form_with url: patient_health_card_path(@patient, format: 'pdf'), class: 'button is-info is-light', method: :post do |f| %>        
```

This should fix the spacing issue:

Screen Shot 2021-05-14 at 1 17 21 PM

radamson:
Overall looks pretty good to me. It would be nice if clicking the `Download PDF` button also disabled the other buttons too since you can actually click something like the `Edit` button after and go to another page before the form post finishes before being redirected to the PDF. I'm not sure the best way to handle this from a UX perspective (is it just using javascript to disable the other buttons?)

app/views/patients/show.html.erb
mickohanlon23: Thanks!

mickohanlon23:
> > 1. I actually try to download the PDF to my disk in Chrome, I receive a "Failed - Network error" message. However, I am able to properly download when using Safari. Are people having the same issue on their machines?
> 
> I see this too. It looks like when you click the download down arrow that chrome is actually trying to access the resource directly from the server
> 
> Screen Shot 2021-05-14 at 1 04 29 PM
> 
> using the Chrome DevTools I see `Failed to load resource: net::ERR_FILE_NOT_FOUND`. Im guessing chrome is actually trying to access this file but its not available from a `GET http://server.com/patients/1/health_card.pdf`, only from a `POST http://server.com/patients/1/health_card.pdf` with form data that includes the qr code. You can work around the download issue by doing a Print > Save as PDF, but it's not a real replace meant. (And I'm not exactly sure this is what is going on since when I look in the network tab it doesn't actually seem like a subsequent request is made...)



> > 1. I actually try to download the PDF to my disk in Chrome, I receive a "Failed - Network error" message. However, I am able to properly download when using Safari. Are people having the same issue on their machines?
> 
> I see this too. It looks like when you click the download down arrow that chrome is actually trying to access the resource directly from the server
> 
> Screen Shot 2021-05-14 at 1 04 29 PM
> 
> using the Chrome DevTools I see `Failed to load resource: net::ERR_FILE_NOT_FOUND`. Im guessing chrome is actually trying to access this file but its not available from a `GET http://server.com/patients/1/health_card.pdf`, only from a `POST http://server.com/patients/1/health_card.pdf` with form data that includes the qr code. You can work around the download issue by doing a Print > Save as PDF, but it's not a real replace meant. (And I'm not exactly sure this is what is going on since when I look in the network tab it doesn't actually seem like a subsequent request is made...)



> > 1. I actually try to download the PDF to my disk in Chrome, I receive a "Failed - Network error" message. However, I am able to properly download when using Safari. Are people having the same issue on their machines?
> 
> I see this too. It looks like when you click the download down arrow that chrome is actually trying to access the resource directly from the server
> 
> Screen Shot 2021-05-14 at 1 04 29 PM
> 
> using the Chrome DevTools I see `Failed to load resource: net::ERR_FILE_NOT_FOUND`. Im guessing chrome is actually trying to access this file but its not available from a `GET http://server.com/patients/1/health_card.pdf`, only from a `POST http://server.com/patients/1/health_card.pdf` with form data that includes the qr code. You can work around the download issue by doing a Print > Save as PDF, but it's not a real replace meant. (And I'm not exactly sure this is what is going on since when I look in the network tab it doesn't actually seem like a subsequent request is made...)

I'm not quite sure how to fix this, since I believe the QR generation requires a POST so that the Javascript can be accessed to create it. Any thoughts? I tried adding a GET Request that would more or less do the same thing as the POST, but I am not able to generate the PDF with the QR Code this way.

mickohanlon23:
> > 1. I actually try to download the PDF to my disk in Chrome, I receive a "Failed - Network error" message. However, I am able to properly download when using Safari. Are people having the same issue on their machines?
> 
> I see this too. It looks like when you click the download down arrow that chrome is actually trying to access the resource directly from the server
> 
> Screen Shot 2021-05-14 at 1 04 29 PM
> 
> using the Chrome DevTools I see `Failed to load resource: net::ERR_FILE_NOT_FOUND`. Im guessing chrome is actually trying to access this file but its not available from a `GET http://server.com/patients/1/health_card.pdf`, only from a `POST http://server.com/patients/1/health_card.pdf` with form data that includes the qr code. You can work around the download issue by doing a Print > Save as PDF, but it's not a real replace meant. (And I'm not exactly sure this is what is going on since when I look in the network tab it doesn't actually seem like a subsequent request is made...)

I'm not quite sure how to fix this, since I believe the QR generation requires a POST so that the Javascript can be accessed to create it. Any thoughts? I tried adding a GET Request that would more or less do the same thing as the POST, but I am not able to generate the PDF with the QR Code this way.

mickohanlon23:
> Overall looks pretty good to me. It would be nice if clicking the `Download PDF` button also disabled the other buttons too since you can actually click something like the `Edit` button after and go to another page before the form post finishes before being redirected to the PDF. I'm not sure the best way to handle this from a UX perspective (is it just using javascript to disable the other buttons?)

Dan just pushed a commit to fix this - now you shouldn't be able to click anywhere else on the patient show page after Download PDF is clicked.

ssayer:
As far the GET vs POST issue, you could try changing the form to use GET, but I'm not sure if there's an issue with that. @Jammjammjamm probably knows.

Jammjammjamm:
GETs don't have bodies. The ideal solution is to get qr code generation working in ruby, which requires updating a dependency.

config/routes.rb
```diff
@@ -8,7 +8,7 @@
     resources :immunizations
     resource :health_card do
       get 'chunks', format: :json
-
+      post "/qrcode", to: "health_cards#qrcode", as: :qrcode, format: :json
```
ssayer, line 11: I don't think you need this route, as you're using health_cards#create

ssayer (changes requested):
Needs a controller test for health_cards#create with pdf format

mickohanlon23:
> GETs don't have bodies. The ideal solution is to get qr code generation working in ruby, which requires updating a dependency.

Ticket [DVCI-97](https://jira.mitre.org/browse/DVCI-97) created to resolve this, but I think it may be fixed by [DVCI-101](https://jira.mitre.org/browse/DVCI-101).

ssayer (approved):
Looks good. I'll refactor my ruby QR code branch once this is in, and that will address the Chrome issues.

radamson (approved):
Looks good and does the job.

It would be nice for the other disabled buttons to also be styled in a way that indicates they are disabled (i.e. all buttons should be greyed out, not just "Download PDF"), but I don't want to hold this up for that. 

Screen Shot 2021-06-03 at 9 55 26 AM

It would also be good to put another task in the backlog to make the PDF look a bit nicer if/when html & css versions of the designs are published.

mickohanlon23:
> Looks good and does the job.
> 
> It would be nice for the other disabled buttons to also be styled in a way that indicates they are disabled (i.e. all buttons should be greyed out, not just "Download PDF"), but I don't want to hold this up for that.
> 
> Screen Shot 2021-06-03 at 9 55 26 AM
> 
> It would also be good to put another task in the backlog to make the PDF look a bit nicer if/when html & css versions of the designs are published.

We already have [DVCI-98](https://jira.mitre.org/browse/DVCI-98), which I updated this morning to include the link to the Figma page and the Zulip thread where the reference implementations are being discussed.

PR #44, opened 2021-05-13
SMART on FHIR support

danlee1025:
- Added `AuthController`
    - Handles `authorize` and `token` endpoints
    - Added `AUTH_CODE` and `CLIENT_ID` environment variables. `authorize` endpoint will redirect with `AUTH_CODE` as the code. User can then exchange that code for an access token by calling the `token` endpoint.
    - `token` will exchange code for access token that expires in 1 hour. The response also includes the 1st patient in the patient list.
- Added `SecuredController` that authorizes requests
    - Any request without an `Authorization` header will be authorized. This is to make it easier to test endpoints without getting an access token.
    - Access tokens in the `Authorization` header will be checked for correct signature and that it hasn't expired.
    - Currently `PatientsController` and `HealthCardsController` controller extend `SecuredController` so those endpoints will be checked for authorization.

I was having a lot of trouble with CORS when I was trying to test the standalone launch so I added the `rack-cors` gem. Is there a better way to handle this?

To test:
I forked an existing example-smart-app to test the standalone launch. If you put `https://danlee1025.github.io/smart-on-fhir-tutorial/example-smart-app/launch-patient.html?iss=http://localhost:3000` in your browser the app will go through the SMART authorization steps and display the `Patient` resource.


app/controllers/auth_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+      render json: { errors: ['Unauthorized client_id'] }, status: :unauthorized
+    end
+  end
+
+  def token
+    params = request.parameters
+    if params[:code] == Rails.application.config.auth_code
+      scope = ['launch/patient', 'patient/Immunization.read']
+      payload = { exp: helpers.convert_time_to_epoch(Time.current + 3600), scope: scope }
+      token = JWT.encode payload, Rails.application.config.hc_key.key, 'ES256'
```
ssayer, line 22: You should be able to use our existing HC code and not have to pull in JWT

config/application.rb
```diff
@@ -20,5 +20,11 @@ class Application < Rails::Application
     #
     # config.time_zone = "Central Time (US & Canada)"
     # config.eager_load_paths << Rails.root.join("extras")
+    config.middleware.insert_before 0, Rack::Cors do
```
ssayer, line 23: I'm not a CORS expert, but it seems this is duplicative of the set_cors_header. I think using a library is probably better than the simple implementation I made, but we should pick one way to do CORS and be consistent.

ssayer (changes requested):

app/controllers/auth_controller.rb
ssayer: Oh wait this is a JWT and not a JWS... still with some refactoring we should be able use our existing code to take care of this without JWT. I'll leave that decision to @radamson though because it seems like this is a fairly experimental, testing-only PR.

danlee1025: I was looking into utilizing the `JWS` class in the `HealthCards` library but it felt weird since creating a bearer token really has nothing to do with health cards. Up to you guys though I can go either way.

config/application.rb
danlee1025: I removed `set_cors_header` since the library is already doing it. The library also makes handling the `OPTIONS` routes easier.

ssayer (approved):

app/controllers/auth_controller.rb
radamson: I'm a little torn on this, but I do think it would be nice to use the code we already have just so we don't need to pull in another dependency and so we can eat our own dog food. Luckily, it's a pretty simple encode/decode and conveniently uses the same signing algorithm.

My vote is to use our `JWS` class in `health_cards`, but I could be swayed.

danlee1025: Removed the dependency :)

radamson (approved):
I like how you implemented this using the digital signatures to track whether the token is valid!

I tested out with an invalid client id and got an error on the launch redirect as expected (see screenshot)

<details>
  <summary>Invalid Client ID Screenshot</summary>
  
Screen Shot 2021-06-03 at 4 48 44 PM

</details>

I also did a quick test in Inferno which highlighted a few really minor errors we should address at some point. Namely:

* the token endpoint doesn't fail with an invalid client id
* The `pragma: no-cache` header and `cache-control: no-store` headers are missing on the token response
* the response should be a 400, not a 401 (except for [special cases according the the OAuth RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2))

<details>
  <summary>Inferno Screenshot</summary>

Screen Shot 2021-06-03 at 4 50 56 PM

</details>

These issues are really minor and mostly only apply when a client is doing something wrong so I think this is good as is, but we should address them.

@danlee1025 if you want to address these in this PR that's cool with me, otherwise we should add a ticket.

PR #47, opened 2021-05-17
Small bug fix for QR code display bug on patient show page

danlee1025:
Multiple canvas elements were being added to the qr-code container when using the back and forward buttons on the patient show page. I added logic to clear the container before appending new canvas elements on the page.


mickohanlon23 (approved):
Fixes the issue with the moving QR codes!

PR #51, opened 2021-05-18
Fixed types for VC 

pmadria:
Fixed the types for the VC to be smarthhealth.cards and also added immunization


danlee1025 (approved):

danlee1025:
Before merging to `main` ensure an immunization is present before adding `https://smarthealth.cards#immunization` to types.

PR #52, opened 2021-05-18
Restrict upload to .smart-health-card files.

danlee1025:


mickohanlon23:
Is there a way to add error handling for the situation where the "Upload" button is clicked before the "Choose a file" button? i.e. so that there is not a server error when you try to upload no file.

danlee1025:
@mickohanlon23 Just pushed a change for it, I actually implemented this in `main` but it didn't make it into `connectathon`

mickohanlon23 (approved):

PR #56, opened 2021-05-24
CONNECTATHON - Fixed types for VC

mickohanlon23:
PR from Connectathon.

Discuss with @pmadria for review. Original PR: https://github.com/dvci/health_cards/pull/51.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # HealthCards
 
-Creating verifiable credentials with FHIR using [SMART Health Cards](https://healthwallet.cards/)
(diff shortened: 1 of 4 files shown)


mickohanlon23 (commented):
Again, can't officially approve, but this change makes sense according to https://smarthealth.cards/vocabulary/.

ssayer (changes requested):
This makes sense and matches code I had contributed to #44, but haven't made it in due to other issues with that.

My only request is a test to confirm the new behavior.

danlee1025 (approved):

PR #59, opened 2021-05-24
CONNECTATHON - Change QR video size

mickohanlon23:
PR from CONNECTATHON.

See @danlee1025 for review. Original PR: https://github.com/dvci/health_cards/pull/54.

I created a new branch/commit instead of rebasing since it was just a one line change

### DIFF
--- a/app/views/health_cards/scan.html.erb
+++ b/app/views/health_cards/scan.html.erb
@@ -24,7 +24,7 @@
       <div class="level-right" id="multi-status-container">
       </div>
     </div>
-    <video id="preview"></video>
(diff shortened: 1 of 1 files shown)


mickohanlon23 (commented):
Tough to judge whether the overall sensitivity to QR codes is better, but it still works for me when I scan in a QR code from the spec.

danlee1025:
Yea I'm not sure how useful this is... I haven't felt like it's made scanning easier. We can probably leave this change out unless people think it's made it easier to scan.

radamson:
I didn't see a noticeable improvement when testing at the Connectathon.

PR #60, opened 2021-05-25
Use FHIR Metadata when minifying bundles

ssayer:
Refactor passes existing tests. Putting up as draft for sanity check review while I test new capabilities.

NOTE: This turns the stripped FHIR bundle into an actual FHIR::Bundle instead of a nested Hash.

### DIFF
--- a/lib/health_cards/health_card.rb
+++ b/lib/health_cards/health_card.rb
@@ -144,7 +144,7 @@ module HealthCards
           type: self.class.types,
           credentialSubject: {
             fhirVersion: self.class.fhir_version,
-            fhirBundle: strip_fhir_bundle
+            fhirBundle: strip_fhir_bundle.to_hash
           }
         }
 
--- a/lib/health_cards/health_card.rb
+++ b/lib/health_cards/health_card.rb
@@ -84,7 +84,6 @@ module HealthCards
       def allow(klass, attributes)
         allowable[klass] = attributes
 
-        # keys.map ( key.is_a? klass)
       end
 
       # Define disallowed attributes for this HealthCard class
--- a/lib/health_cards/health_card.rb
+++ b/lib/health_cards/health_card.rb
@@ -56,14 +56,33 @@ module HealthCards
       # @param attributes [Array] An array of string with the attribute names that will be passed through
       #  when data is minimized
       def allow(klass, attributes)
-        resource_type = klass.name.split('::').last
-        allowable[resource_type] = attributes
+        allowable[klass] = attributes
+      end
+
+      # Define disallowed attributes for this HealthCard class
+      # @param klass [Class] Scopes the attributes to a spefic class. Must be a subclass of FHIR::Model
+      # @param attributes [Array] An array of string with the attribute names that will be passed through
+      #  when data is minimized
+      def disallow(klass, attributes)
+        disallowable[klass] = attributes
       end
 
       # Define allowed attributes for this HealthCard class
       # @return [Hash] A hash of FHIR::Model subclasses and attributes that will pass through minimization
       def allowable
-        @allowable ||= {}
+        return @allowable if @allowable
+
+        base = superclass == HealthCards::HealthCard ? superclass.allowable : {}
+        @allowable = base
+      end
+
+      # Define disallowed attributes for this HealthCard class
+      # @return [Hash] A hash of FHIR::Model subclasses and attributes that will pass through minimization
+      def disallowable
+        return @disallowable if @disallowable
+
+        base = superclass == HealthCards::HealthCard ? superclass.disallowable : {}
+        @disallowable = base
       end
 
       # Sets/Gets the fhir version that will be passed through to the credential created by an instnace of
@@ -99,6 +118,10 @@ module HealthCards
         !types.intersection(type).empty?
       end
     end
+
+    allow FHIR::Meta, %w[security]
+    disallow FHIR::CodeableConcept, %w[text]
+    disallow FHIR::Coding, %w[display]
 
     # Create a HealthCard
     #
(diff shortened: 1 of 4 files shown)


lib/health_cards/health_card.rb
```diff
@@ -146,78 +169,90 @@ def to_json(*_args)
...
-          member_element.delete('display') if attribute_name == 'coding'
-          update_nested_elements(member_element) if member_element.is_a?(Hash)
-        end
-      end
+    def process_reference(url_map, ref)
+      url = ref.reference
+
+      return unless url
+
+      new_url = url_map.key?(url) ? url_map[url] : url_map["#{issuer}/#{url}"]
```
mickohanlon23, line 251: When is the url map going to contain something of the format "issuer/url"?

test/health_cards/health_card_test.rb
```diff
@@ -21,6 +21,10 @@ class HealthCardTest < ActiveSupport::TestCase
     assert @health_card.bundle.is_a?(FHIR::Bundle)
   end
 
+  # test "metadata walking" do
+  #   byebug
+  # end
```
mickohanlon23, line 26: Do we want to keep this?

test/health_cards/health_card_test.rb
```diff
@@ -95,44 +99,66 @@ class HealthCardTest < ActiveSupport::TestCase
...
 
   test 'update_nested_elements populates Reference.reference elements with short resource-scheme URIs' do
     stripped_bundle = @health_card.strip_fhir_bundle
-    stripped_resources = stripped_bundle['entry']
+    stripped_resources = stripped_bundle.entry
     resource_with_reference = stripped_resources[2]
 
-    reference = resource_with_reference['resource']['subject']['reference']
+    reference = resource_with_reference.resource.subject.reference
     assert(reference.start_with?('resource:') && (reference.length <= 12))
```
mickohanlon23, line 154: This is just a question, but do you know if there is a definition somewhere for "short resource-scheme URIs"? I think the motivation for checking if length was less than 12 was that it can contain "resource:" + a 2 digit number, but I'm realizing now that this test is somewhat arbitrary.

lib/health_cards/health_card.rb
ssayer: This allows us to match FHIR references such as "Patient/1" (which is how we create them in our FHIR code). Issuer here is a url so it would be something like "http://example.com/Patient/1". That actually came from a previous PR, it just got moved here.

test/health_cards/health_card_test.rb
ssayer: Oops, I was originally going to write new tests but decided to just make it pass the existing ones to start. I'll remove this

test/health_cards/health_card_test.rb
ssayer: I don't see it in the spec anywhere, not sure if @masnick or @radamson know. We could change the test to use a regex instead of the existing assertion.

ssayer: I changed this test to use a regex but it would be good to know either way if there's any other rules around this url

mickohanlon23: Let's bring it up at standup next week.

mickohanlon23 (approved):
Refactor makes sense. Looks good! I like the new method of updating/stripping the resource elements.

lib/health_cards/health_card.rb
radamson: I'm not sure this is quite right since the issuer (which is really just an indication of the endpoint hosting the JWKS) might not be the same as the FHIR endpoint where the FHIR data came from.

If we look at the [FHIR docs for resolving references in a bundle](https://www.hl7.org/fhir/bundle.html#references) we really should "extract the [root] from the fullUrl, and append the reference (type/id) to it". In this case the fullUrI is from the resource containing the reference.

radamson (commented):
Looks good.

Lots of side effects with `handle_allowable`, `handle_disallowable`, and `redfine_uris`. These are all private methods so I think it makes sense to keep the code less verbose (and maybe more ruby-y). I don't think it has to change, but I'd be interested to learn from your expert developer wisdom and opinions regarding this @ssayer.

lib/health_cards/health_card.rb
```diff
@@ -146,78 +169,90 @@ def to_json(*_args)
...
+      url_map = redefine_uris(new_bundle)
+
+      walk_resource(new_bundle) do |value, type|
+        case type
+        when 'Reference'
+          value.reference = process_reference(url_map, value)
+        when 'Resource'
+          value.id = nil
+          value.text = nil
+          value.meta = nil unless value.meta&.security
```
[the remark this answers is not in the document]
ssayer: It could be modified to do that, but as it exists now allow/disallow can't handle these cases. My thought was the implementation of these special cases wasn't worth it in the short term at least:

* The allow/disallow rules are based on the class, so I'd have to change it to also use ancestors (e.g. `disallow FHIR::Model, &w(id text)`
* Meta is a special case where you only nil it out if a condition is met. I'd have to add the ability to pass a proc to to allow/disallow to determine if the rule should be applied.

lib/health_cards/health_card.rb
```diff
@@ -146,78 +169,90 @@ def to_json(*_args)
...
     def strip_fhir_bundle
-      stripped_bundle = @bundle.to_hash
-      if stripped_bundle.key?('entry') && !stripped_bundle['entry'].empty?
-        entries = stripped_bundle['entry']
-        entries, @url_map = redefine_uris(entries)
-        update_elements(entries)
+      return [] unless bundle.entry
+
+      # Using this to dup the original bundle, hopefully it correctly dups everything
+      new_bundle = FHIR::Bundle.new(bundle.to_hash)
```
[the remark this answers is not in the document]
ssayer: I investigated deep_dup. Deep_dup for an object just calls dup, so it seemed like this would be good enough. It's easy enough to externalize this into a method though.

lib/health_cards/health_card.rb
```diff
@@ -100,6 +119,10 @@ def supports_type?(*type)
       end
     end
 
+    allow FHIR::Meta, %w[security]
+    disallow FHIR::CodeableConcept, %w[text]
+    disallow FHIR::Coding, %w[display]
+
```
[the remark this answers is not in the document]
ssayer: This wouldn't work (unless I'm misreading one of your suggested commits) as allow/disallow doesn't consider ancestors for determining if an attribute is allowed or disallowed. It just looks at the class of the resource.

ssayer:
@radamson For some reason I can't seem to directly reply to your comment about the urls. This feature isn't for building urls, this is for matching existing urls. When we have an Immunization that references a patient with "Patient/1" we need to be able to match that to a fullUrl when we minify our bundles. This feature allows us to match the partial reference that we hold in our FHIR JSON to a fullUrl in a bundle. I can add an option to pass in any url instead of forcing the use of the issuer url if you want. This works for our implementation, but you're right it might not work for everyone.

test/health_cards/health_card_test.rb
```diff
@@ -95,44 +95,67 @@ class HealthCardTest < ActiveSupport::TestCase
...
+
+    # Check entry value
+    patient = stripped_bundle.entry[0].resource
+    patient.name[0].family = 'foobar'
+    assert_not_equal @health_card.bundle.entry[0].resource.name[0].family,
+                     stripped_bundle.entry[0].resource.name[0].family
+
+    # Check entries
+    stripped_bundle.entry = []
+    assert_not_equal @health_card.bundle.entry.length, stripped_bundle.entry.length
```
Jammjammjamm, line 113: I think a decent test here would be to convert the bundle to a JSON string, strip the bundle, convert the original bundle to a JSON string again and make sure they're the same.

lib/health_cards/health_card.rb
```diff
@@ -56,14 +56,33 @@ def compress_payload(payload)
       # @param attributes [Array] An array of string with the attribute names that will be passed through
       #  when data is minimized
       def allow(klass, attributes)
-        resource_type = klass.name.split('::').last
-        allowable[resource_type] = attributes
+        allowable[klass] = attributes
```
Jammjammjamm, line 59: Since we permit inheriting these, should `allow/disallow` concatenate rather than overwrite?

If so, I'd initialize them with `Hash.new { |hash, key| hash[key] = [] }` rather than just `{}`, and then update `allow/disallow` to be `[dis]allowable[klass].concat(attributes)`.

radamson: +1 for supporting concatenation

ssayer: I think this is a good idea, but I'd say it's out of scope for this PR. I can create a ticket with a few of the suggested improvements (concatenation of allow/disallow rules, use inheritance in allow/disallow rules) and work on that later.

lib/health_cards/health_card.rb
```diff
@@ -56,14 +82,33 @@ def compress_payload(payload)
...
+
+        # keys.map ( key.is_a? klass)
```
radamson, line 87:
```suggestion
```

lib/health_cards/health_card.rb
```diff
@@ -98,8 +143,22 @@ def types
       def supports_type?(*type)
         !types.intersection(type).empty?
       end
+
+      protected
+
+      def parent_allowables(base = {})
+        self < HealthCards::HealthCard ? superclass.parent_allowables(base.merge(superclass.allowable)) : base
```
radamson, line 150: `superclass.allowable` should already contain `superclass.parent_allowables`

```suggestion
        self < HealthCards::HealthCard ? base.merge(superclass.allowable) : base
```

lib/health_cards/health_card.rb
```diff
@@ -98,8 +143,22 @@ def types
...
+      end
+
+      def parent_disallowables(base = {})
+        self < HealthCards::HealthCard ? superclass.parent_disallowables(base.merge(superclass.disallowable)) : base
```
radamson, line 154: `superclass.disallowable` should already contain `superclass.parent_disallowables`

```suggestion
        self < HealthCards::HealthCard ? base.merge(superclass.disallowable) : base
```

radamson (approved):
Big improvements here to how the minification is done.

Would be great to see some dedicated tests for the `allow`/`disallow` functionality (maybe I just missed where they were)

ssayer:
I'll make a note in the ticket to improve testing on those features as a part of the upgrades

PR #62, opened 2021-05-27
QR code generation in Ruby

ssayer:
Server side generation of QR codes in Ruby. Since a JWS may resolve to multiple QR codes you need to supply an ordinal to generate one of the QR codes. The images are available using the url scheme '/patients/:patient_id/qr/:ordinal.png'


danlee1025:
It doesn't look like it decodes correctly when I scan the 1st qr code into msft verifier portal https://demo-portals.smarthealth.cards/VerifierPortal.html

image

Are there some extra zeroes padded at the beginning?

ssayer:
> It doesn't look like it decodes correctly when I scan the 1st qr code into msft verifier portal https://demo-portals.smarthealth.cards/VerifierPortal.html
> 
> image
> 
> Are there some extra zeroes padded at the beginning?

Possible. Thanks for letting me know... I'll reviewing the numeric encoding

mickohanlon23 (changes requested):

app/views/health_cards/show.pdf.erb
```diff
@@ -55,8 +55,16 @@
...
     </div>
-    <div id="qr-code"></div>
-    <img src = '<%=@image_uri%>' >
-
+    <div id="qr-code">
+      <div class="tile is-ancestor">
+        <% @exporter.qr_codes.chunks.each do |ch| %>
+          <div class="tile is-parent">
+            <div class="tile is-child box has-text-centered">
+              <%= wicked_pdf_image_tag patient_qr_code_url(@patient, ch.ordinal, format: :png) %>
```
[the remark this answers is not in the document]
ssayer: I shrunk the QR code and it can now support up to 3 immunizations without paging. We may want to consider adding a task to redesign the PDF as each immunization takes up a lot of space, and I'm not sure if we should have all the patient data displayed either.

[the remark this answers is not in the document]
mickohanlon23: I agree. [DVCI-98](https://jira.mitre.org/browse/DVCI-98) should take care of this. @ssayer feel free to add anything to the ticket that you think should be modified with the patient data.

Gemfile
```diff
@@ -22,6 +22,8 @@ gem 'webpacker', '~> 5.0'
 gem 'turbolinks', '~> 5'
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
 gem 'jbuilder', '~> 2.7'
+
+gem 'rqrcode'
```
Jammjammjamm, line 26: I'd like to see us require the `health_cards` gem (just pointed at `path: '.'`) rather than adding its dependencies to the app's Gemfile. I didn't notice anywhere where we were using `rqrcode` outside of the library, but if we are then you can ignore this comment.

app/controllers/auth_controller.rb
```diff
@@ -16,14 +16,15 @@ def authorize
   def token
     params = request.parameters
     if params[:code] == Rails.application.config.auth_code
+      ttl = 1.hour
```
Jammjammjamm, line 19: If we're bothering to improve the readability of `3600`, I think we should use an unabbreviated variable name.

Jammjammjamm (approved):
This worked well for me.

Gemfile
ssayer: Removed json-minify and rqrcode from Gemfile and added health_cards reference

radamson (approved):
Looks good. Tested a newly generated QR against the `validation-sdk` without issue (ignore the https issue - I was testing on localhost)

```shell
SMART Health Card Validation SDK v1.0.0-3

QR image
   │  
   └─ QR numeric
         │  
         └─ JWS-compact
               │  
               ├─ Error
               │    · Issuer URL SHALL use https
               │  
               └─ JWS.payload
                     │  
                     └─ FhirBundle

Validation completed
```

PR #63, opened 2021-06-09
Better Error Handling in HealthCards library and FHIR endponits

ssayer:
The code was using params[:format] to check if the request was FHIR JSON. This only works if you use the .format url format that works in Rails (but is not aligned with the FHIR spec). Changed to use the more rails-y respond_to which works in both cases. Added test to confirm this.


app/controllers/application_controller.rb
```diff
@@ -7,10 +7,13 @@ def create_exporter
     @patient = Patient.find(params[:patient_id])
     @exporter = COVIDHealthCardExporter.new(@patient)
   rescue ActiveRecord::RecordNotFound => e
-    raise e unless params[:format] == 'fhir_json'
-
-    issue = FHIR::OperationOutcome::Issue.new(severity: 'error', code: 'not-found',
-                                              diagnostic: 'Patient does not exist')
-    render json: FHIR::OperationOutcome.new(issue: issue).to_json, status: :not_found and return
+    respond_to do |format|
```
Jammjammjamm, line 10: The patients controller needs this same handling. Performing a FHIR read on a non-existent patient returns html. Same for immunizations.

ssayer: Roger, I'll add that too

lib/health_cards/exporter.rb
```diff
@@ -14,14 +14,39 @@ def download(jws)
       # Export JWS for $issue-health-card endpoint
       # @param [Array<JWS>, String] An array of JWS objects to be exported
       # @return [String] JSON string containing a FHIR Parameters resource
-      def issue(jws)
-        params = jws.map { |j| FHIR::Parameters::Parameter.new(name: 'verifiableCredential', valueString: j) }
+      def issue(fhir_params)
```
Jammjammjamm, line 17: The documentation for this method needs updating. It doesn't look like it takes a string anymore.

lib/health_cards/health_card.rb
```diff
@@ -138,7 +138,7 @@ def types
       # Check if this class supports the given type claim(s)
       # @param type [Array, String] A type as defined by the SMART Health Cards framework
       # @return [Boolean] Whether or not the type param is included in the types supported by the HealthCard (sub)class
-      def supports_type?(*type)
+      def supports_type?(type)
```
Jammjammjamm, line 141: The documentation says this method can take a string, but it looks to me like it will only work with an array.

lib/health_cards/exporter.rb
```diff
@@ -14,14 +14,39 @@ def download(jws)
...
         QRCodes.from_jws(jws)
       end
+
+      private
+
+      def extract_types!(fhir_params)
+        if fhir_params.nil?
+          code = 'structure'
+          err_msg = 'No Parameters found'
+        elsif !fhir_params.valid?
```
Jammjammjamm, line 35: I don't think the validation in fhir models is robust enough that we should rely on it in the library.

app/controllers/application_controller.rb
```diff
@@ -1,16 +1,40 @@
...
+  end
+
+  def find_patient
     @patient = Patient.find(params[:patient_id])
-    @exporter = COVIDHealthCardExporter.new(@patient)
-  rescue ActiveRecord::RecordNotFound => e
-    raise e unless params[:format] == 'fhir_json'
+  end
+
+  def handle_fhir_errors
```
Jammjammjamm, line 16: This is nice.

lib/health_cards/exporter.rb
ssayer: I looked this up and the only validation it does is check that each FHIR::Parameter has a populated name attribute. I could code that manually or remove this line if you prefer but I don't think it's going to cause any issues as is.

lib/health_cards/health_card.rb
ssayer: Fix code so that it accepts either

lib/health_cards/exporter.rb
ssayer: fixed

app/controllers/application_controller.rb
ssayer: Thanks! I wasn't sure if there were other errors we should specially handle that wouldn't result in a 500 error.

PR #65, opened 2021-06-10
WIP: CodeCov Debug

radamson:
Do not merge. Trying to debug codecov.


radamson:
Seems like the status checks are working now and its using the correct code-cov bot.

I have no idea what thing I actually did to make this work, but I think maybe we just had to be more patient after enabling? Also, I don't think any configuration should be done on codecov.io (i.e. only enable in GH and don't touch the config in codecov.io). the only thing I did in codecov.io was delete the repo, but never enabled anything there. 🤷

PR #66, opened 2021-06-10
DVCI-118: Improve site navigation

Jammjammjamm:
This branch ensures that each page has a banner at the top which includes a back button. This will generally take you up the page hierarchy (patient show -> patient list -> landing page, etc.). There are a few pages which are accessible from multiple places, so these link to `:back`, which uses javascript to look at your history and go back a page.


ssayer (approved):

radamson (commented):
Tested it out and it worked well.

One thing I did notice is that the message for a successful patient creation or update appears above the banner. Should it be below it? Seems wrong to have it above, but what do I know

Screen Shot 2021-06-16 at 3 21 51 PM

radamson:
Any reason to not use `:back` in every case? One downside I see is that we can no longer can populate the `aria-label`, but it seems like it would be less brittle.

Jammjammjamm:
Yes, because we don't always want to go back. For instance, I want the back button on the patient show page to go to the patient list, not the new patient form.

Jammjammjamm:
> One thing I did notice is that the message for a successful patient creation or update appears above the banner. Should it be below it? Seems wrong to have it above, but what do I know

That should be an easy fix.

radamson (approved):

PR #67, opened 2021-06-16
Refactor/Improve Allowed/Disallowed Attributes

ssayer:
**Features/Changes**

1. Moved some HealthCard subclass specific behavior into modules to make HealthCard more readable
2. Disallowed now inherits from parent class. So if a parent health card class disallows attribute A on a resource and the subclass disallows attribute B, then the subclass will actually disallow both A and B. This functionality was not implemented for allowable because it's likely there would be conflicts between a parent and child subclass.
3. Updated API for allow/disallow so parameters are now named. Omitting a type: parameter will result in a rule applying for every resource and type.

We don't have an actual example yet for for Feature #2, not sure how to better test this.


lib/health_cards/attribute_filters.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+module HealthCards
+  # Handles behavior related to removing disallowed attributes from FHIR Resources
+  module AttributeFilters
+    def self.included(base)
+      base.send :include, InstanceMethods
```
Jammjammjamm, line 7: I'd get rid of the `InstanceMethods` submodule and just stick those in `AttributeFilters`, since it is already being included.

lib/health_cards/attribute_filters.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      def parent_disallowables(base = {})
+        self < HealthCards::HealthCard ? base.merge(superclass.disallowable) : base
+      end
+    end
+
+    # Disallow attributes on resource instances
+    module InstanceMethods
+      def handle_allowable(resource)
+        # byebug if resource.is_a?(FHIR::Patient)
```
Jammjammjamm, line 60: Remove comment.

Jammjammjamm (approved):

PR #68, opened 2021-06-16
Dvci 116 fix token responses

Shaumik-Ashraf:
Addresses [this comment](https://github.com/dvci/health_cards/pull/44#pullrequestreview-675761570) from a previous PR.

I changed error response code 401 to 400 and differentiated between invalid_client and invalid_request in `auth_controller.rb`

I set the headers Pragma: no-cache and Cache-Control: no-store (in auth_controller only)

I turned caching on in the test environment and added a Last-Modified header to auth_controller responses. This is because rails was otherwise overwriting the Cache-Control header.

I also updated the tests in `auth_controller_test.rb` to reflect these changes.


app/controllers/auth_controller.rb
```diff
@@ -26,10 +29,20 @@ def token
         token_type: 'Bearer',
         expires_in: time_to_live,
         scope: scope,
-        patient: Patient.all.first.id
+        patient: Patient.first&.id
       }
+    elsif params.key? :code
```
danlee1025, line 34: We also need to check `:client_id` in `def token`. `code` and `client_id` are two separate params.

danlee1025 (commented):
This looks really good! Just needs one more check.

test/controllers/auth_controller_test.rb
```diff
@@ -19,15 +23,37 @@ class HealthCardsControllerTest < ActionDispatch::IntegrationTest
...
+    post(auth_token_path, params: { code: 'bad_code' })
+    assert_response :bad_request
+    assert_no_cache_headers
+    assert_equal('{"error":"invalid_client"}', @response.body)
   end
 
   test 'token with correct code' do
     post(auth_token_path, params: { code: Rails.application.config.auth_code })
     assert_response :success
+    assert_no_cache_headers
```
mickohanlon23, line 46: Should we make a separate test that calls `assert_no_cache_headers`? Right now if something is wrong with the headers, every test will fail, which may make it slightly hard to debug from a unit test perspective. I don't think there is benefit to adding the two header asserts to every single test.

test/controllers/auth_controller_test.rb
```diff
@@ -2,15 +2,19 @@
...
-    assert_response :unauthorized
+    assert_response :bad_request
+    assert_no_cache_headers
+    assert_equal('{"error":"invalid_request"}', @response.body)
   end
 
-  test 'authorize with incorrect client_id should return 401' do
-    get(auth_code_path, params: { client_id: 'id' })
-    assert_response :unauthorized
+  test 'authorize with incorrect client_id should return 400 invalid_client' do
```
mickohanlon23, line 13: Is there ever a situation where this will return a 401 instead of a 400, according to https://datatracker.ietf.org/doc/html/rfc6749#section-5.2?

test/controllers/auth_controller_test.rb
```diff
@@ -19,15 +23,37 @@ class HealthCardsControllerTest < ActionDispatch::IntegrationTest
...
+  end
+
+  test 'token with incorrect code should return 400 invalid_client' do
+    post(auth_token_path, params: { code: 'bad_code' })
+    assert_response :bad_request
+    assert_no_cache_headers
+    assert_equal('{"error":"invalid_client"}', @response.body)
   end
 
   test 'token with correct code' do
```
mickohanlon23, line 43: Kind of nit-picky, but can we add what it "should" do to the title of this test since all of the other tests are formatted that way now.

app/controllers/auth_controller.rb
Shaumik-Ashraf: I changed it so it will be:
```ruby
elsif params.key? :code or params.key? :client_id
      (return 400 bad_request with error: invalid_client)
else
      (return 400 bad_request with error: invalid_request)
```
but is this correct? To my understanding client_id should only be passed to /auth/authorize, and if its sent to /auth/token the error message should be error: invalid_request

Shaumik-Ashraf: (those return statements are pseudo-code btw)

test/controllers/auth_controller_test.rb
Shaumik-Ashraf: I considered that, but then I would either make one test that tests all action methods for the correct headers (which then makes it ambiguous which method caused the error) or make individual tests for each method just to test the header (which seems redundant). Should I do the latter anyways?

test/controllers/auth_controller_test.rb
Shaumik-Ashraf: Fixed

test/controllers/auth_controller_test.rb
Shaumik-Ashraf: I think that should be in the SecuredController, when client sends an Authorize http header but the bearer token is incorrect a 401 must be returned. In the AuthController the client shouldn't be sending any Authorization header, just a GET /auth/authorize?client_id=something.

Shaumik-Ashraf: But in SecuredController right now we just accept incorrect Authorization headers for development purposes I think.

app/controllers/auth_controller.rb
```diff
@@ -26,10 +29,20 @@ def token
         token_type: 'Bearer',
         expires_in: time_to_live,
         scope: scope,
-        patient: Patient.all.first.id
+        patient: Patient.first&.id
       }
+    elsif params.key?(:code) || params.key?(:client_id)
```
Jammjammjamm, line 34: I believe this is a public app workflow, so client id is required in the token exchange.

http://hl7.org/fhir/smart-app-launch/#step-3-app-exchanges-authorization-code-for-access-token

test/controllers/auth_controller_test.rb
Jammjammjamm: I'm with Mick on this one. It's only two controller actions, so one test for each action checking the cache headers rather than checking them in every test.

config/environments/test.rb
```diff
@@ -26,8 +26,8 @@
 
   # Show full error reports and disable caching.
   config.consider_all_requests_local       = true
-  config.action_controller.perform_caching = false
-  config.cache_store = :null_store
+  config.action_controller.perform_caching = true            # originally false
+  config.cache_store = :memory_store, { size: 64.megabytes } # originally :null_store
```
Jammjammjamm, line 30: Enabling caching in all tests makes me nervous. But if we really want to do this, no need to have comments with the previous values. We can always look at the git history if we care.

test/controllers/auth_controller_test.rb
mickohanlon23: In the case of "one test that tests all action methods for the correct headers," wouldn't an error in the correct headers be the fault of something within the `assert_no_cache_headers` function, and as a result propagate to _all_ of the action methods? If that is the case, I don't think there would be a situation where you would need to decipher which action method caused the errors, and you could just throw all the actions into one test.

If not, I think it might be worth being redundant to modularize the test cases, but it's probably not too big of a deal.

test/controllers/auth_controller_test.rb
mickohanlon23: Makes sense!

app/controllers/auth_controller.rb
danlee1025: Like Steve said below since this is the [public app workflow](http://hl7.org/fhir/smart-app-launch/#step-3-app-exchanges-authorization-code-for-access-token), both `code` and `client_id` are required. So we should check if `client_id` is equal to `Rails.application.config.client_id`, same check in `/auth/authorize`

test/controllers/auth_controller_test.rb
Jammjammjamm: The fact that both of the requests get their cache headers are set in the same way is an implementation detail that the tests don't need to know about. Also, having a separate test for each endpoint makes it more likely that we'll remember to implement that test if we add another action to the controller (such as a token refresh).

I would not modularize the tests or do anything fancy at all with them. The simpler a test is the better.

app/controllers/auth_controller.rb
Shaumik-Ashraf: In that case should we also check the client_id before handing the token? I.e:
```ruby
if params[:code] == <auth_code> && params[:client_id] == <client id>
    render json: <token stuff>
```

config/environments/test.rb
Shaumik-Ashraf: Removed the comments. Have to turn caching on since the test env will otherwise overwrite the cache-control header. I don't think there is anyway to just turn caching on for one test.

Shaumik-Ashraf: Also, I fixed the tests as Mick suggested now.

test/controllers/auth_controller_test.rb
Shaumik-Ashraf: I fixed this now, doing a separate test for the headers.

app/controllers/auth_controller.rb
Shaumik-Ashraf: ^^ done

app/controllers/auth_controller.rb
Shaumik-Ashraf: got it

app/controllers/auth_controller.rb
```diff
@@ -26,10 +29,20 @@ def token
         token_type: 'Bearer',
         expires_in: time_to_live,
         scope: scope,
-        patient: Patient.all.first.id
+        patient: Patient.first&.id
       }
+    elsif params.key?(:code) && params.key?(:client_id)
+      render json: { error: 'invalid_client' }, status: :bad_request
     else
-      render_operation_outcome(code: 'unknown', http: :unauthorized, message: 'Unauthorized Code')
```
danlee1025, line 32: Looks like Sam had the `token` endpoint return an operation outcome so we should probably keep that here instead of just rendering json.

Shaumik-Ashraf: I asked him about that, there seems to a conflict between the OAuth spec and the FHIR spec, my implementations following the OAuth spec so he just said to keep it

app/controllers/auth_controller.rb
```diff
@@ -26,10 +29,20 @@ def token
         token_type: 'Bearer',
         expires_in: time_to_live,
         scope: scope,
-        patient: Patient.all.first.id
+        patient: Patient.first&.id
       }
+    elsif params.key?(:code) && params.key?(:client_id)
```
mickohanlon23, line 34: Should a correct `client_id` with an incorrect `code` return an `invalid_grant` instead of an `invalid_client`?

https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

Shaumik-Ashraf: Yes, I'm adding that case

Shaumik-Ashraf: I'm also adding a check for `grant_type` param; if `grant_type` **is** included but does not equal `authorization_code` I return `invalid_grant_type`

OAuth 2 actually requires `grant_type` but I'm keeping it optional: <https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3>

mickohanlon23: I think we should add a test for that case, if you're adding the `elsif` statement that checks for an invalid grant type.

mickohanlon23: And is there a reason you're keeping it optional? it seems like grant_type is required according to [7.1.3](http://hl7.org/fhir/smart-app-launch/#step-3-app-exchanges-authorization-code-for-access-token)

Shaumik-Ashraf: I mean I'm keeping the param grant_type=authorization_code optional, since in the FHIR spec grant_type is not listed as a required param

Jammjammjamm: I don't know where you're seeing that `grant_type` isn't required. Mick linked to the section that says it is required.
Screen Shot 2021-06-21 at 1 40 20 PM

test/controllers/auth_controller_test.rb
```diff
@@ -2,32 +2,108 @@
...
+  test 'token with no parameters should return 400 invalid_request' do
+    post(auth_token_path)
+    assert_response :bad_request
+    assert_equal('{"error":"invalid_request"}', @response.body)
   end
 
-  test 'token with correct code' do
-    post(auth_token_path, params: { code: Rails.application.config.auth_code })
+  test 'token with missing client_id parameter should return 400 invalid_request' do
+    post(auth_token_path, params: { code: Rails.application.config.client_id })
```
Jammjammjamm, line 36: With these params, it's difficult to reason about exactly what is causing the errors. I think it would be better to have a set of good params, and then modify them in each test so that the parameters are good except for the specific ones that need to be bad for a particular test. Like:
```
post(auth_token_path, params: good_params.merge(code: 'BAD_CODE'))
```

app/controllers/auth_controller.rb
```diff
@@ -3,33 +3,77 @@
...
-        access_token: jws.to_s,
-        token_type: 'Bearer',
-        expires_in: time_to_live,
-        scope: scope,
-        patient: Patient.all.first.id
-      }
+    valid_request      = proc { |p| valid_request?(p)      }
+    invalid_grant      = proc { |p| invalid_grant?(p)      }
+    invalid_client     = proc { |p| invalid_client?(p)     }
+    invalid_grant_type = proc { |p| invalid_grant_type?(p) }
```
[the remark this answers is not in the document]
Shaumik-Ashraf: You're right, I did all that to make it pass rubocop, but the simplest solution is to use private methods the way they're meant to be used. I'm fixing this in the next commit.

app/controllers/auth_controller.rb
```diff
@@ -3,33 +3,77 @@
...
+  def valid_request?(params)
+    params[:client_id] == Rails.application.config.client_id && params[:code] == Rails.application.config.auth_code
+  end
+
+  def invalid_grant?(params)
+    params[:client_id] && params[:code] && params[:code] != Rails.application.config.auth_code
+  end
+
+  def invalid_client?(params)
+    params[:client_id] && params[:client_id] != Rails.application.config.client_id && params[:code]
```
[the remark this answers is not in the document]
Shaumik-Ashraf: Not having `code` should be an invalid request for `auth/token`, but an incorrect `code` should be invalid grant which I check next. Come to think of it, the method should also return `true` if called in `auth/authorize` without `code` so I'll fix that.

app/controllers/auth_controller.rb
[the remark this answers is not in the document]
Jammjammjamm: We can always disable a rubocop rule in a particular location when necessary.

test/controllers/auth_controller_test.rb
Shaumik-Ashraf: I will also add this to next commit.

app/controllers/auth_controller.rb
Shaumik-Ashraf: Ah its only for the /auth/token action. My bad I'm fixing this.

Shaumik-Ashraf:
Since it looks like we are doing a strict implementation instead of SMART app launch instead of minimal, do we want to leverage the [oauth2 gem](https://github.com/oauth-xx/oauth2/tree/v1.4.7)?

Jammjammjamm (approved):

mickohanlon23 (approved):
I like the refactor with the added checks for all the params. Tests now show all the functionality working properly. Great work!

PR #76, opened 2021-06-25
DVCI-133: Align patient form with VCI IG

ssayer:
*Email and Phone Number no longer required
*Only one of given or family name required
*Patient.given will read from name.given or name.text

This is a draft until #75 is in, since I need to fix how we handle text elements in non-Resource types during the walk.

### DIFF
--- a/lib/health_cards/attribute_filters.rb
+++ b/lib/health_cards/attribute_filters.rb
@@ -3,6 +3,8 @@
 module HealthCards
   # Handles behavior related to removing disallowed attributes from FHIR Resources
   module AttributeFilters
+    ALL_FHIR_RESOURCES = :fhir_resource
+
     def self.included(base)
       base.extend ClassMethods
     end
@@ -13,9 +15,7 @@ module HealthCards
       # @param type [Class] Scopes the attributes to a spefic class. Must be a subclass of FHIR::Model
       # @param attributes [Array] An array of string with the attribute names that will be passed through
       #  when data is minimized
-      def allow(type: nil, attributes: [])
-        raise ArgumentError, 'Must specify a base type for allowable attributes' if type.nil?
-
+      def allow(type:, attributes: [])
         allowable[type] = attributes
       end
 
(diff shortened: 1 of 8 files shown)


lib/health_cards/attribute_filters.rb
```diff
@@ -10,18 +10,21 @@ def self.included(base)
     # Class level methods for HealthCard class specific settings
     module ClassMethods
       # Define allowed attributes for this HealthCard class
-      # @param klass [Class] Scopes the attributes to a spefic class. Must be a subclass of FHIR::Model
+      # @param type [Class] Scopes the attributes to a spefic class. Must be a subclass of FHIR::Model
       # @param attributes [Array] An array of string with the attribute names that will be passed through
       #  when data is minimized
-      def allow(type: FHIR::Model, attributes: [])
+      def allow(type: nil, attributes: [])
+        raise ArgumentError, 'Must specify a base type for allowable attributes' if type.nil?
```
Jammjammjamm, line 17: Is there a reason not to just remove the default to enforce this requirement?

lib/health_cards/attribute_filters.rb
```diff
@@ -10,18 +10,21 @@ def self.included(base)
...
       end
 
       # Define disallowed attributes for this HealthCard class
-      # @param klass [Class] Scopes the attributes to a spefic class. Must be a subclass of FHIR::Model
+      # @param type [Class] Scopes the attributes to a spefic class. If not used will default to all FHIR resources.
+      # To apply a rule to all FHIR types (resources and types), use FHIR::Model as the type
       # @param attributes [Array] An array of string with the attribute names that will be passed through
       #  when data is minimized
-      def disallow(type: FHIR::Model, attributes: [])
+      def disallow(type: nil, attributes: [])
```
Jammjammjamm, line 27: So, leaving this blank applies it to all FHIR resources, and using `FHIR::Model` applies it to resources and elements? It's weird that `nil` is the value for that, but I guess it would also be weird if it were a magic string or something rather than a class.

What do you think about defining a constant in this class and using that as the default value?

Jammjammjamm (approved):

PR #81, opened 2021-07-12
Create IIS consumer portal interface

Shaumik-Ashraf:
- create `search_controller.rb` that handles `GET /search/form` and `POST /search/query`
 - create `views/search/form.html.erb` that has terribly designed input form
 - create `search_helper.rb` that maps above input fields to QBP client query
 - add methods `Patient#match?` and `Patient.create_from_bundle!`
 - add lots of test code (bypasses QBP client by accepting `param[:qbp_response]` in test env)


app/controllers/search_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+  # POST /search/query
+  def query
+    results = qbp_query(query_params)
+
+    case results[:code]
+    when 'AE'
+      redirect_to search_form_url,
+                  alert: 'Sorry there was an error, please try again. If problem persists contact system administrator.'
+    when 'AR'
+      render 'rejected', status: :bad_request
```
Jammjammjamm, line 18: I feel like all of these results other than `OK` could be handled by redirecting back to the search form with a flash message rather than with custom error pages.

app/models/patient.rb
```diff
@@ -133,6 +133,39 @@ def to_bundle(base_url)
     bundle
   end
 
+  # compares model's fhir attributes to the key-value pairs in hash
+  def match?(hash)
+    matching = true
+    hash.each do |k, v|
+      if has_attribute? k
+        matching &&= (send(k).to_s == v.to_s)
+        break unless matching
```
Jammjammjamm, line 142: Rather than using `each` with `break`:
```
hash
  .select { |key, _| has_attribute? key }
  .all? { |key, value| send(key).to_s == value.to_s }
```
Is this really robust for complex types?

app/helpers/patients_helper.rb
```diff
@@ -29,4 +29,11 @@ def fake_patient_params
                  birth_date: birth_date, phone: phone, email: email, street_line1: street_line1,
                  street_line2: street_line2, city: city, state: state, zip_code: zip_code } }
   end
+
+  def iis_patient_params
```
Jammjammjamm, line 33: Is there a reason for this to be here rather than in the test code?

app/models/vaccine.rb
```diff
@@ -1,7 +1,26 @@
 # frozen_string_literal: true
 
 class Vaccine < ApplicationRecord
+  CVX = 'http://hl7.org/fhir/sid/cvx'
+
   has_many :immunizations, dependent: :restrict_with_exception
 
   default_scope { order(:name) }
+
+  def self.seed
```
Jammjammjamm, line 10: I wouldn't change how the seeding works since it's being worked on in another branch.

app/controllers/search_controller.rb
Shaumik-Ashraf: fixed

app/models/patient.rb
Shaumik-Ashraf: I think using `===` is better for complex types but then rubocop fails the code, however using `==` directly fails in functionality. Not sure if there are any other suggestions.

app/helpers/patients_helper.rb
Shaumik-Ashraf: This is for a demo data button I made on the search form, copying the new patient form. I just moved it to the search helper but doesn't really belong in the test code. We also need to ensure this matches a patient in the IIS sandbox so I don't want to throw it in the controller.

app/models/patient.rb
Jammjammjamm: I just did some testing and it seems like fhir models compares complex types at least somewhat intelligently, so we should be fine with `==`.

app/models/vaccine.rb
Shaumik-Ashraf: I needed to seed Vaccine for the test code; since the tests are run in parallel I end up seeding multiple times and decided to put it into one function. I can't use test fixtures because YAML parses the CVX code as an Integer, while its suppose to be a string.

app/helpers/search_helper.rb
```diff
@@ -0,0 +1,84 @@
...
+  # Omits the following QBP parameters from query:
+  # - query[:address][:address_type]
+  # - query[:multiple_birth_indicator]
+  # - query[:birth_order]
+  # - query[:client_last_updated_date]
+  # - query[:client_last_updated_facility]
+
+  FHIR_TO_HL7_SEX = { 'male' => 'M', 'female' => 'F', 'other' => 'U', 'unknown' => 'U' }.freeze
+
+  def sanitize_input(action_params)
```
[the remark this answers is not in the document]
Shaumik-Ashraf: I removed `sanitize_inputs` from search helper and do sanitization in the search controller now. I kept build_query in the helper however since I want all the logic for transforming the forms data into the QBP client input in one place. I do also have a `query_params` function in search controller that ties it all together.

Shaumik-Ashraf:
Closing because this project will be forked into another repository.

PR #86, opened 2021-08-25
Fix Dockerfile and Add GitHub Action

radamson:
## Summary

Fixes the broken Dockerfile so that the Docker image can be built and adds a GitHub Action to detect future regressions.

The key fix, identified by @Jammjammjamm, is that the project files need to be added (using `ADD . $INSTALL_PATH`) before the bundle install is performed. This is because the `health_cards` gem is sourced locally and _not_ from rubygems.org. Thus we need the `health_cards.gemspec` to identify it's dependencies and the `health_cards` gem files (from `spec.files` in the `health_cards.gemspec`) in order to install the `health_cards` gem.

This is less than optimal since we're no longer benefiting as much from the [Docker build cache](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache) and any source code changes will now result in dependencies being re-downloaded on rebuild (instead of being able to leverage the existing cache!) and we expect source will change more frequently than dependencies.

## Testing
- _Optional: Clear your Docker cache (I normally use `docker system prune -a`)_
- Build the Docker image with `docker build -t health_cards .`
- Run the container to make sure it works (e.g. `docker run -p 3000:3000 health_cards` and then access at `http://0.0.0.0:3000/` in a browser)


Dockerfile
```diff
@@ -22,8 +22,10 @@ RUN yarn install
 
 ADD Gemfile* $INSTALL_PATH
 RUN gem install bundler
-RUN bundle install --deployment --without development --without test
+RUN bundle config set --local deployment 'true'
```
mickohanlon23, line 25: The deployment version run by this container throws a 500 error when you attempt to download the PDF for a SMART Health Card.

mickohanlon23 (changes requested):

Dockerfile
mickohanlon23: Let's track this in [DVCI-161](https://jira.mitre.org/browse/DVCI-161).

mickohanlon23 (approved):
I realized that my comment about the PDF Download being broken was most likely an issue before this PR.

The Docker image builds correctly, and the app is fully functional when running via the Docker container, except for the PDF functionality.

I moved my comment about the PDF generation over to [DVCI-161](https://jira.mitre.org/browse/DVCI-161), where we are also tracking a similar issue relating to the PDF generation. Figured we should prioritize merging this PR in order to expedite the AWS deployment. Looks good!

Issue #84, opened 2021-08-13
additional_types aren't inherited


jmandel:
e.g. given https://github.com/dvci/health_cards/blob/4eaa1a5d522fb895a8e15ee188749e1b3129cd02/lib/health_cards/covid_health_card.rb#L11-L12

... the code at https://github.com/HL7/fhir-shc-vaccination-ig/blob/8f1fc04e2f0c81c6b12ea5a2a85189cd5333e849/examples/generate-examples.rb#L6 inadvertently drops the `#covid19` type, so it's missing in https://github.com/HL7/fhir-shc-vaccination-ig/blob/master/examples/Scenario1Bundle-jws-payload-expanded.json#L5-L8

jmandel:
Same issue applies for fhir_version.

nkarip:
We're tracking this issue now internally in our project Jira, will update here as needed.

ssayer:
@nkarip @jmandel The changes to allow inheritance of these attributes are now in master, and should fix this issue. If you can confirm all your issues are addressed, please close this ticket. A few implementation notes below:

- We have set 4.0.1 as the base FHIR version for all health cards. This can be overridden by subclasses, but based on the Zulip topic this seemed to make sense.
- `COVIDImmunizationCard` should now be used instead of `COVIDHealthCard`. We were previously including all types in `COVIDHealthCard`, which I don't think is the expected behavior. Support for actually generating lab cards from the app is still forthcoming.

radamson:
I'm going to close this ticket since @ssayer addressed it in https://github.com/dvci/health_cards/pull/87, but if anyone is still seeing issues related to this feel free to reopen it!

PR #87, opened 2021-08-27
Make fhir_version and types inheritable

ssayer:
Fix for #84. Makes fhir_version and types inheritable by HealthCard subclasses (and subclasses of subclasses).

Note: While HealthCards::HealthCard has a base type defined (https://smarthealth.cards#health-card), it does not define a fhir version. I'm not sure if we should, but as is if you create a HealthCard it will have a null fhir_version. You'll have to use a subclass in order to create a valid HealthCard.


test/health_cards/covid_health_card_test.rb
```diff
@@ -8,20 +8,27 @@ class COVIDHealthCardTest < ActiveSupport::TestCase
...
   end
 
   test 'includes required credential attributes in hash' do
     hash = @card.to_hash
     type = hash.dig(:vc, :type)
     assert_not_nil type
-    assert_includes type, HealthCards::CardTypes::VC_TYPE[0]
+    assert_includes type, 'https://smarthealth.cards#health-card'
     assert_includes type, 'https://smarthealth.cards#covid19'
     assert_includes type, 'https://smarthealth.cards#immunization'
```
mickohanlon23, line 33: Should we implement types so that `#immunization` and `#laboratory` only appear if those resources are present, or is @danlee1025 covering that in [DVCI-87](https://jira.mitre.org/browse/DVCI-87)?

Right now, a patient bundle with no lab nor immunization data still includes these types, which throws warnings when I test with the Microsoft Verifier portal.

mickohanlon23:
> Fix for #84. Makes fhir_version and types inheritable by HealthCard subclasses (and subclasses of subclasses).
> 
> Note: While HealthCards::HealthCard has a base type defined (https://smarthealth.cards#health-card), it does not define a fhir version. I'm not sure if we should, but as is if you create a HealthCard it will have a null fhir_version. You'll have to use a subclass in order to create a valid HealthCard.

Is it even possible to create a non-subclassed health card with our current reference implementation? I wasn't able to create a VC that doesn't include a FHIR version.

test/health_cards/covid_health_card_test.rb
```diff
@@ -8,20 +8,27 @@ class COVIDHealthCardTest < ActiveSupport::TestCase
...
     assert_includes type, 'https://smarthealth.cards#labresult'
```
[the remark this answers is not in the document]
ssayer: I'll make the change here

test/health_cards/health_card_test.rb
```diff
@@ -3,6 +3,13 @@
 require 'test_helper'
 
 class HealthCardTest < ActiveSupport::TestCase
+  class TestCOVIDHealthCard < HealthCards::COVIDHealthCard
```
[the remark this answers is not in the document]
ssayer: I'll delete this. I was originally going to test the inheritance here, but I decided to do it in COVIDHealthCard test in order to ensure that a subclass of a subclass also worked correctly.

test/health_cards/covid_health_card_test.rb
ssayer: I think that's out of scope for this PR, as I'm only trying to address inheritance issues, but it would make sense for DVCI-87

ssayer:
> Additionally, would the Vaccination & Testing IG being based upon FHIR 4.0.1 be enough to declare that as the default FHIR version?

I posted on Zulip for feedback and I feel comfortable enough now to make 4.0.1 the "default". This won't prevent users from subclassing and going back to DSTU3 if they really want.

test/health_cards/covid_health_card_test.rb
ssayer: Actually, on second thought since theoretically this we would end up with invalid HC examples if I leave it as is, I might revisit this after I talk to @danlee1025.

ssayer:
It seems like there's a need to create un-filtered healthcards, so I'm going to add that feature here while I'm at it since the focus here is on generating examples.

ssayer:
> It seems like there's a need to create un-filtered healthcards, so I'm going to add that feature here while I'm at it since the focus here is on generating examples.

After discussing with @mickohanlon23 offline this will go into a different PR

app/controllers/health_cards_controller.rb
```diff
@@ -13,7 +13,7 @@ def show
       format.healthcard { render json: exporter.download }
       format.html do
         @jws_encoded_details = exporter.jws
-        @health_card = HealthCards::COVIDHealthCard.from_jws @jws_encoded_details
+        @health_card = HealthCards::COVIDImmunizationCard.from_jws @jws_encoded_details
```
mickohanlon23, line 16: Right now, we can only create `COVIDImmunizationCard` classes, and not `COVIDLabResultCard` classes. This seems wrong to not be able to have the `#laboratory` types in VCs that contain only lab data. 

Is this being addressed in [DVCI-87](https://jira.mitre.org/browse/DVCI-87)?

mickohanlon23 (approved):
This addresses all the problems brought up in Issue #84, so I'm approving. 

I left one additional comment, but I'm realizing now that it is probably way beyond the scope of this PR and better addressed by @danlee1025's current task.

Nice job.

app/controllers/health_cards_controller.rb
danlee1025: Yep it's being addressed in my task!

PR #88, opened 2021-09-07
Multi Mode Support - Non Hacky Implementation

ssayer:
`rqrcode_core` 1.2.0 has been released with support for multi segment encoding. This PR moves us over to use that version, and eliminates the hacky implementation we currently have. This will also support smaller QR codes as we currently hard code to version 22.


mickohanlon23 (approved):
QR Code renders correctly and validates when scanned. Also seems to scan quicker with our verifier app, which is nice. Added bonus of eliminating a console warning we were previously getting with the PDF download. Looks great 👍

radamson (approved):
Did a `rails assets:clobber`, before testing (just to be sure we weren't caching anything).

Worked great and code is way cleaner now. Hopefully a future release of `rqrcode` embeds the updated `rqrcode_core` version eventually so we don't have to explicitly have that dependency requirement forever.

The QR code did look a lot smaller too (I didn't test the QR code version, but I'm guessing it was under 22).

qr

Results from https://github.com/smart-on-fhir/health-cards-dev-tools also indicate the QR stuff is working right:


```shell
SMART Health Card Dev Tools v1.0.2-0

QR image
   │  
   └─ QR numeric
         │  
         └─ JWS-compact
               │  
               ├─ Error
               │    · Issuer URL SHALL use https
               │  
               └─ JWS.payload
                     │  
                     ├─ Warning
                     │    · JWS.payload.vc.type SHOULD NOT contain 'https://smarthealth.cards#immunization', no immunization resources found
                     │    · JWS.payload.vc.type SHOULD NOT contain 'https://smarthealth.cards#covid19', no covid immunization or observation found
                     │  
                     └─ FhirBundle
```

PR #91, opened 2021-09-10
Allow PDF to render in production using Docker

mickohanlon23:
Adding the webpacker commands to the Dockerfile allowed me to render the Downloaded PDF when running the app using Docker. Can somebody confirm that this also works on their machine?

Note: I had to delete some dated React files, since webpacker didn't like interpreting jsx files. It looks like these files were unused anyways.

I also added the necessary Docker commands to the README. Those can be used for testing.


radamson:
I'm able to get the PDF to render using Docker on `main`, but I probably had recently clobbered locally.

`clobber` deletes the files in `health_cards/public/packs`, which isn't in our `.dockerignore`. We could add `/public/packs` to `.dockerignore` instead which might be slightly cleaner from a process perspective (i.e. instead of copying over the compiled assets and then deleting them, just don't move them at all). That being said, I think I might prefer the `clobber` method since that mimics what we're doing locally. Eventually, a comprehensive fix for docker and local development would be great and it will be easier to identify this as something that could be removed once we figure that out.

Dockerfile
```diff
@@ -31,6 +31,9 @@ ENV KEY_PATH=config/keys/key.pem
 ENV SECRET_KEY_BASE=f7c9be19114730b947c8f7f274ea7c128e792245d049ae2c808d479f7e632817dd83009fa235326953a9ab11c70e9b0be4a8cb8657626e57ca5031a584b75295
 ENV RAILS_SERVE_STATIC_FILES=true
 
+RUN bundle exec rake webpacker:clobber
+RUN bundle exec rake webpacker:install
```
radamson, line 35: Do we really need `webpacker:install`? Based on the documentation this seems like something that we should only have to [do once to configure Rails to use webpacker](https://github.com/rails/webpacker#installation).

README.md
```diff
@@ -41,6 +42,15 @@ Run server:
 
 Then go to `http://127.0.0.1:3000` to view the locally running application.
 
+#### Docker:
+
+Alternatively, you can create a Docker image and start it in a container:
+```
+docker build -t health_cards .
+docker run -p 3000:3000 health_cards
```
radamson, line 50: I should have added this in my PR, but can you update this to include the necessary environment variable:

```suggestion
docker run -p 3000:3000 --env HOST=http://localhost:3000 health_cards
```

mickohanlon23: I added this to the command and then added a sentence right after instructing issuers to pass in their own personal HOST environment variable.

Dockerfile
mickohanlon23: I don't think it's needed. I tried rebuilding the image without the command and everything seemed to work fine, however, I'm not sure if things would be different if I cleared my cache. I'll remove it.

mickohanlon23:
> I'm able to get the PDF to render using Docker on `main`, but I probably had recently clobbered locally.
> 
> `clobber` deletes the files in `health_cards/public/packs`, which isn't in our `.dockerignore`. We could add `/public/packs` to `.dockerignore` instead which might be slightly cleaner from a process perspective (i.e. instead of copying over the compiled assets and then deleting them, just don't move them at all). That being said, I think I might prefer the `clobber` method since that mimics what we're doing locally. Eventually, a comprehensive fix for docker and local development would be great and it will be easier to identify this as something that could be removed once we figure that out.
> 
> Do we know what types of changes require a `webpacker:clobber`? It would be super useful to know a simple change to make that would show that the assets haven't been properly recompiled in order to test a more comprehensive fix.

I don't personally. @ssayer Do you have any ideas?

danlee1025:
Looks good and I'm able to get the pdf to render.

I get this error when scanning a QR code from https://spec.smarthealth.cards/examples/ 
image

Unsure if this is just a problem with my machine or a docker issue. If it is a docker issue is this something we need to fix?

radamson (approved):

ssayer:
> > I'm able to get the PDF to render using Docker on `main`, but I probably had recently clobbered locally.
> > `clobber` deletes the files in `health_cards/public/packs`, which isn't in our `.dockerignore`. We could add `/public/packs` to `.dockerignore` instead which might be slightly cleaner from a process perspective (i.e. instead of copying over the compiled assets and then deleting them, just don't move them at all). That being said, I think I might prefer the `clobber` method since that mimics what we're doing locally. Eventually, a comprehensive fix for docker and local development would be great and it will be easier to identify this as something that could be removed once we figure that out.
> > Do we know what types of changes require a `webpacker:clobber`? It would be super useful to know a simple change to make that would show that the assets haven't been properly recompiled in order to test a more comprehensive fix.
> 
> I don't personally. @ssayer Do you have any ideas?

@mickohanlon23 I'm not sure. I think we have two choices:

1. Add the packs dir to our .gitignore and .dockerignore (thought that might be redundant) to make dev environment just like prod.
2. Just use the clobber/compile everywhere

ssayer:
> Unsure if this is just a problem with my machine or a docker issue. If it is a docker issue is this something we need to fix?

It's a network issue with your SSL certs, it shouldn't affect people who use the docker image.

danlee1025 (approved):

mickohanlon23:
> > > I'm able to get the PDF to render using Docker on `main`, but I probably had recently clobbered locally.
> > > `clobber` deletes the files in `health_cards/public/packs`, which isn't in our `.dockerignore`. We could add `/public/packs` to `.dockerignore` instead which might be slightly cleaner from a process perspective (i.e. instead of copying over the compiled assets and then deleting them, just don't move them at all). That being said, I think I might prefer the `clobber` method since that mimics what we're doing locally. Eventually, a comprehensive fix for docker and local development would be great and it will be easier to identify this as something that could be removed once we figure that out.
> > > Do we know what types of changes require a `webpacker:clobber`? It would be super useful to know a simple change to make that would show that the assets haven't been properly recompiled in order to test a more comprehensive fix.
> > 
> > 
> > I don't personally. @ssayer Do you have any ideas?
> 
> @mickohanlon23 I'm not sure. I think we have two choices:
> 
> 1. Add the packs dir to our .gitignore and .dockerignore (thought that might be redundant) to make dev environment just like prod.
> 2. Just use the clobber/compile everywhere

It looks like we somewhat do Choice #2 currently, so I'll keep things as-is for now since there are already 2 approvals. Happy to submit a follow-up PR if we decide on better practice for handling packs going forward.

mickohanlon23:
> > > I'm able to get the PDF to render using Docker on `main`, but I probably had recently clobbered locally.
> > > `clobber` deletes the files in `health_cards/public/packs`, which isn't in our `.dockerignore`. We could add `/public/packs` to `.dockerignore` instead which might be slightly cleaner from a process perspective (i.e. instead of copying over the compiled assets and then deleting them, just don't move them at all). That being said, I think I might prefer the `clobber` method since that mimics what we're doing locally. Eventually, a comprehensive fix for docker and local development would be great and it will be easier to identify this as something that could be removed once we figure that out.
> > > Do we know what types of changes require a `webpacker:clobber`? It would be super useful to know a simple change to make that would show that the assets haven't been properly recompiled in order to test a more comprehensive fix.
> > 
> > 
> > I don't personally. @ssayer Do you have any ideas?
> 
> @mickohanlon23 I'm not sure. I think we have two choices:
> 
> 1. Add the packs dir to our .gitignore and .dockerignore (thought that might be redundant) to make dev environment just like prod.
> 2. Just use the clobber/compile everywhere

It looks like we somewhat do Choice #2 currently, so I'll keep things as-is for now since there are already 2 approvals. Happy to submit a follow-up PR if we decide on better practice for handling packs going forward.

PR #92, opened 2021-09-13
debug qr codes

ssayer:
* Gracefully handle unknown Vaccine codes by creating an "Unknown Vaccine" entry in the database. This isn't the optimal solution but will prevent the system from bombing out on scan for the Connectathon.
* Adds new vaccines from the accepted CDC list
* Adds a script to inspect QR code payloads on the command line
Usage: 

`./test_app/test_qr_codes shc:2342343...`


db/seeds.rb
```diff
@@ -15,6 +15,14 @@
...
   vaccine.doses_required = 2
 end
+Vaccine.find_or_create_by(code: '210') do |vaccine|
+  vaccine.name = 'AstraZeneca COVID-19 Vaccine'
+  vaccine.doses_required = 2
+end
+Vaccine.find_or_create_by(code: '211') do |vaccine|
+  vaccine.name = 'Novavax COVID-19 Vaccine'
+  vaccine.doses_required = 2
+end
```
mickohanlon23, line 25: Are we purposefully we leaving out the Non-US Covid vaccines in 
https://www.cdc.gov/vaccines/programs/iis/COVID-19-related-codes.html#fact-sheets

mickohanlon23 (approved):
Nice! I tested with the CO Implementation code, and a card for an Unknown Vaccine came up as expected.

Was able to run the script with an example from the spec, and it seemed to log all the expected QR code information. 

I left one question about the vaccines we're supporting, but technically speaking looks good!

db/seeds.rb
ssayer: I was originally going to do that, but after talking to @masnick I'll update this to include them all

ssayer: Discussed this on the track this morning, and I'm including all with a trade name. We'll make it more complete when we refactor

db/seeds.rb
```diff
@@ -15,7 +15,28 @@
...
+  vaccine.name = 'Sinopharm (BIBP) COVID-19 Vaccine'
+  vaccine.doses_required = 2
+end
+Vaccine.find_or_create_by(code: '511') do |vaccine|
+  vaccine.name = 'Coronavac (Sinovac) COVID-19 Vaccine'
+  vaccine.doses_required = 2
+end
+
+Vaccine.find_or_create_by(code: '500') do |vaccine|
+  vaccine.name = 'Coronavac (Sinovac) COVID-19 Vaccine'
```
mickohanlon23, line 40: Why is `500` also listed as 'Coronavac (Sinovac) COVID-19 Vaccine'? According to https://www.cdc.gov/vaccines/programs/iis/COVID-19-related-codes.html, doesn't 500 indicate "Product Unknown"?

ssayer: That's weird, I must not have hit save. Pushing a fix for this now.

Issue #94, opened 2021-09-16
patient page causes error if the patient was modified


gjskha:
If you create a patient, then go to any given patient page at the `/patients/N` route, where N is the patient ID, it will display fine. Then if you click "Record Immunization" or "Record Lab Result", once it is recorded, when the app redirects back to the `/patients/N` route, you will get an error with a backtrace:

```
ArgumentError (bad argument (expected URI object or URI string)):
lib/health_cards/health_card.rb:202:in `process_reference'
lib/health_cards/health_card.rb:161:in `block (2 levels) in strip_fhir_bundle'
lib/health_cards/health_card.rb:158:in `block in strip_fhir_bundle'
lib/health_cards/health_card.rb:157:in `each'
lib/health_cards/health_card.rb:157:in `strip_fhir_bundle'
lib/health_cards/health_card.rb:128:in `to_hash'
lib/health_cards/health_card.rb:144:in `to_json'
lib/health_cards/health_card.rb:138:in `to_s'
lib/health_cards/issuer.rb:35:in `issue_jws'
app/lib/patient_exporter.rb:36:in `jws'
app/lib/patient_exporter.rb:13:in `qr_codes'
app/controllers/patients_controller.rb:15:in `block (2 levels) in show'
app/controllers/patients_controller.rb:14:in `show'
app/controllers/application_controller.rb:31:in `handle_fhir_errors'
```

However, if you go to the json version e.g. `/patients/N.json` it renders fine. So it appears to be a problem with the HTML rendering.

gjskha:
The triggering line looks different from all other format.html calls in that file. Its the first one below.
```
     format.html { @qr_codes = exporter.qr_codes }
        format.html { redirect_to @patient, notice: 'Patient was successfully created.' }
        format.html { render :new, status: :unprocessable_entity }
        format.html { redirect_to @patient, notice: 'Patient was successfully updated.' }
        format.html { render :edit, status: :unprocessable_entity }
      format.html { redirect_to patients_url, notice: 'Patient was successfully destroyed.' }
```

ssayer:
Hi @gjskha, I'm having trouble reproducing this locally, but judging from the stack trace the issue seems to be caused by some kind of URL processing issue when creating the health card. This is done when displaying the QR code, which is only done in the HTML version.

It looks like you might be using an old version though, can you update to the latest version of master and let me know if you still have the issue? Also, have you made any modification to the .env files?

ssayer:
[...]

Issue #89, opened 2021-09-07
PDFs don't render at "/patients/[patient id]/health_card.pdf"


gjskha:
PDFs don't render when fetching the PDF at the route "/patients/[number]/health_card.pdf" where [number] is the patient's ID. I found this message in the logs:

```
ActionView::Template::Error (The asset "qr-code.js" is not present in the asset pipeline.
):
1: <%= javascript_pack_tag 'qrcode' %>
2: <%= wicked_pdf_javascript_include_tag "qr-code" %>
3:
4: <%=wicked_pdf_stylesheet_link_tag 'pdf' %>
5: <p id="notice"><%= notice %></p>

app/views/health_cards/show.pdf.erb:2
app/controllers/health_cards_controller.rb:21:in `block (2 levels) in show'
app/controllers/health_cards_controller.rb:12:in `show'
app/controllers/application_controller.rb:31:in `handle_fhir_errors'
```

I noticed that there is a file at "app/javascript/packs/qrcode.js" so I changed  the erb at line 2 above from ```qr-code``` to ```qrcode``` but it doesn't make a difference -- looks like it could be doing some fancy caching somewhere.

ssayer:
PR #88 should address this, as it finalizes our migration from JS based QR code generation to server side.

radamson:
@gjskha can you update to the latest `main` branch and see if the recent merge of #88 has fixed your issue? I'm able to get the PDF to render on main locally and in docker.

If you are using docker you'll need to set the host appropriately (e.g. `docker run -p 3000:3000 --env HOST=http://localhost:3000 health_cards`)

gjskha:
I can indeed generate PDFs now, using non-Docker method.  Closing.

PR #97, opened 2021-09-17
Add Nginx

radamson:
Adds nginx to support deploying behind a load balancer that does SSL termination. Nginx is configured to provide the `X-Forwarded` type headers to rails which address `ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken):` errors.

**References:**
* https://github.com/rails/rails/issues/22965


radamson:
I forgot to update the README with instructions on how to run. I'll do that shortly.

mickohanlon23:
Mention reference this PR in Issue #89 after merging.

radamson:
Check that `iis` is populated when running locally, with docker, and with docker compose.

mickohanlon23:
> Check that `iis` is populated when running locally, with docker, and with docker compose.

Confirmed that `iss` is populated to `http://localhost:3000` when running all 3 ways.

app/views/health_cards/show.pdf.erb
```diff
@@ -87,7 +87,9 @@
         <% @qr_codes.chunks.each do |ch| %>
           <div class="tile is-parent">
             <div class="tile is-child box has-text-centered">
-              <%= wicked_pdf_image_tag patient_qr_code_url(@patient, ch.ordinal, format: :png) %>
+              <%= wicked_pdf_image_tag url_for(controller: 'qr_codes', action: 'show', host: "localhost",
+                                               protocol: 'http', port: ENV['PORT'], patient_id: @patient.id,
```
mickohanlon23, line 91: It looks like this causes the QR Code in the PDF to not render when running using Docker (it still works fine when running with docker-compose).

I was able to fix this by adding `ENV PORT=3000` to `Dockerfile`.

mickohanlon23 (changes requested):

app/views/health_cards/show.pdf.erb
radamson: Good catch. Fixed in https://github.com/dvci/health_cards/pull/97/commits/42ad67f0a9ee6791cc17a9194ec458ecbd3f6349.

mickohanlon23: QR Code properly renders now!

mickohanlon23 (approved):
All functionality works when running locally, using Docker, and using Docker-compose. 

Most recent commit seems to successfully address Issue #94. I tested passing in different HOST environment variables to the docker container, as well as not passing in any, and neither scenario produced errors. 👍

PR #99, opened 2021-11-17
Health Cards API Redesign

ssayer:
This separates the usage and creation of a health card into two different classes: HealthCard and Payload

* `HealthCards::HealthCard` becomes `HealthCards::Payload`
* A new `HealthCard::HealthCard` class which includes data needed to use/share a health card (e.g. jws, qr codes, bundle).
* This also addresses the issue with re-issuing health cards (i.e. bad QR codes) by putting the current health card in the session.

I need to do more testing/refinement, but posting as a draft for feedback on major API changes


lib/USAGE.md
```diff
@@ -32,26 +32,48 @@ jws = 'foofoofoo.barbarbar.bazbazbaz'
...
 verifier.verify(jws)
+
 ```
 
-## Manually Operation
+## Creating/Using a Health Card
 
-Most applications will want to use the library from `Issuer` or `Verifier`.
-Health Cards can also be manually created and verified if more control is needed.
+HealthCards encapsulate a JWS in order to faciliate the use of the data by systems
```
radamson, line 40:
```suggestion
HealthCards encapsulate a data in a JWS in order to faciliate the use of the data by systems
```

lib/USAGE.md
```diff
@@ -64,20 +86,24 @@ jws.public_key = public_key
 
 # JWS will verify using the included public_key
 jws.verify
+
+health_card = HealthCards::HealthCard.new(jws)
+health_card.verify
```
radamson, line 91: Does this method actually work? Verifying requires a public key which `Verifier`'s have, but HealthCards themselves don't.

We'd at least need to have `verify` accept a key or keyset if we wanted to keep this functionality.

lib/health_cards.rb
```diff
@@ -9,17 +9,18 @@
...
 require 'health_cards/qr_codes'
 require 'health_cards/chunk'
 require 'health_cards/verifier'
 require 'health_cards/importer'
-require 'health_cards/covid_health_card'
-require 'health_cards/covid_immunization_card'
-require 'health_cards/covid_lab_result_card'
+require 'health_cards/covid_payload'
+require 'health_cards/covid_immunization_payload'
+require 'health_cards/covid_lab_result_payload'
```
radamson, line 23: Should we have a `payload_types` subdirectory so that we can just import `health_cards/payload_types`? We have a decent amount of library files so it might be a good time to start grouping files together.

lib/health_cards/health_card.rb
```diff
@@ -1,214 +1,58 @@
...
-      def fhir_version(ver = nil)
-        if @fhir_version.nil? && ver.nil?
-          @fhir_version = superclass.fhir_version unless self == HealthCards::HealthCard
-        elsif ver
-          @fhir_version = ver
-        end
-        @fhir_version
-      end
+    def initialize(jws)
+      @jws = jws.is_a?(String) ? JWS.from_jws(jws) : jws
```
radamson, line 15:
```suggestion
      @jws = JWS.from_jws(jws)
```

`JWS.from_jws` gracefully handles the JWS type (String, HealthCards::JWS, etc.).

lib/health_cards/health_card.rb
```diff
@@ -1,214 +1,58 @@
...
-    #
-    # @param bundle [FHIR::Bundle] VerifiableCredential containing a fhir bundle
-    # @param issuer [String] The url from the Issuer of the HealthCard
-    def initialize(bundle:, issuer: nil)
-      raise InvalidPayloadError unless bundle.is_a?(FHIR::Bundle) # && bundle.valid?
-
-      @issuer = issuer
-      @bundle = bundle
+    def credential
+      { verifiableCredential: [@jws.to_s] }
```
radamson, line 21:
```suggestion
    def download
      Exporter.download(@jws)
```

This is already implemented in [Exporter where it is called `download`](https://github.com/dvci/health_cards/blob/bfb1d66811662270f777222ae50aeb3502a876d4/lib/health_cards/exporter.rb#L10-L12) since it's actually the [File Download Format](https://spec.smarthealth.cards/#via-file-download).

We should be consistent with the naming. I don't really like the `download` name either though so I'm up for suggestions (`file_download_json`?)

lib/health_cards/health_card.rb
```diff
@@ -1,214 +1,58 @@
...
-        vc: {
-          type: self.class.types,
-          credentialSubject: {
-            fhirVersion: self.class.fhir_version,
-            fhirBundle: fhir_bundle.to_hash
-          }
-        }
-      }
+    def to_json(*_args)
+      credential.to_json
```
radamson, line 25: If we change `credential` we'll need to update the name here too (and in other places)

radamson (commented):
I think this is a great change (which is probably not a surpise). Breaking up `HealthCard` into `HealthCard` and `Payload` helps break up the logic nicely and makes it simpler to reason with (HealthCards are signed clinical data, and if you change the data or signature you have a new Health Card).

I really like the `HealthCards.resource` and `HealthCards.resouces` functionality.

Left a couple small suggestions for your consideration.

CHANGELOG.md
```diff
@@ -12,7 +12,7 @@ _Note: This changelog is tracking changes related to the `health_cards` library.
 - Updated Bundle minification to use `each_element` [\#75](https://github.com/dvci/health_cards/pull/75)
 - Updated README [\#70](https://github.com/dvci/health_cards/pull/70)
 - Removed `json-minify` dependency [\#69](https://github.com/dvci/health_cards/pull/69)
-- Fixed `COVIDHealthCard` VC Type [\#56](https://github.com/dvci/health_cards/pull/56)
+- Fixed `COVIDPayload` VC Type [\#56](https://github.com/dvci/health_cards/pull/56)
```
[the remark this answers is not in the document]
ssayer: This was a find/replace error on my part. I'll revert it.

lib/USAGE.md
ssayer: We discussed this offline, but basically it delegates to the internal JWS. If the the JWS is from an external source (and we don't have the source public key), it'll use remote key resolution. Otherwise, it'll use the key contained in the JWS object.

I'll update `Verifier` so that it can accept a `HealthCard`, in addition to a `JWS`

lib/health_cards.rb
ssayer: Yes, I think this is a good idea. At some point we may want to consider splitting out the COVID specific stuff into its own gem so it can be updated independently of the underlying SHC implementation.

radamson (approved):

Issue #101, opened 2021-12-30
USAGE.md have mistake


kossy0701:
# Summary

thanks to all contributors.

I try to use health_cards gem reference to [USAGE.md](https://github.com/dvci/health_cards/blob/main/lib/USAGE.md).

but USAGE.md have one mistake.

```
# In console
> key = HealthCards::Key.generate_key
=> (irb):6:in `<main>': undefined method `generate_key' for HealthCards::Key:Class (NoMethodError)

# maybe correct using
> HealthCards::PrivateKey.generate_key
=> #<HealthCards::PrivateKey:0x0000561fa5e3ff88 @key=#<OpenSSL::PKey::EC:0x0000561fa5e34020 oid=id-ecPublicKey>>
```

# FIx Proposal

## before

```
# Generate or load a private key
key = HealthCards::Key.generate_key # or `key = HealthCards::Key.from_file`
```

## After

```
# Generate or load a private key
key = HealthCards::PrivateKey.generate_key # or `key = HealthCards::PrivateKey.from_file`
```

radamson:
@kossy0701 Thanks for catching and reporting this!

This is fixed by https://github.com/dvci/health_cards/pull/102. Take a look and if it resolves the issue feel free to close this issue.

radamson:
Closing this issue since a fix has been implemented in https://github.com/dvci/health_cards/pull/102. Please reopen this issue if you see that this isn't fully addressed now!

PR #102, opened 2022-01-25
Upgrade usage docs

mickohanlon23:
As pointed out in Issue #101, some parts of [lib/Usage.md](https://github.com/dvci/health_cards/blob/main/lib/USAGE.md) are outdated. This pull request updates instructions in Usage.md to align with the most recent refactor of the health card library. This PR also fixes several formatting inconsistencies.

Steps within Usage.md were tested using `test_app/test_health_cards_without_rails`


lib/USAGE.md
```diff
@@ -101,21 +100,21 @@ Custom `Payload` or `Issuer` class can be created to customize their behavior.
...
 
 **Should the global configuration be prefixed with `globally_`? e.g. `HealthCards.globally_resolve_keys = false`**
 
 #### Globally
 Public key resolution can be configured across the library. This will affect the public key resolution
-for all classes and instances, including `Issuer` and `Payload
+for all classes and instances, including `Issuer` and `Payload`
 ```ruby
 # The current state can be checked with:
 HealthCards.resolves_keys?
```
mickohanlon23, line 120: I was unable to get both `#resolves_keys?` and `#resolve_keys=` working for both `HealthCards` and `HealthCards::Verifier`. Any ideas for the best way to resolve this @radamson?

radamson: `#resolve_keys` and `resolve_keys` are only `HealthCards::Verifier` instance methods. Setting key resolution at library and class level was something we were thinking about when `HealthCards` were responsible for more. So we can delete most of this except

```
Public key resolution can be disabled for a single instance issuer with:
```ruby
verifier = HealthCards::Verifier.new
# The current state can be checked with:
verifier.resolve_keys?
# Public key resolution can be disabled with
verifier.resolve_keys = false
# Keys can be manually added that the verifier can use to verify credentials
key = HealthCards::PrivateKey.from_file('my_keys.pem')
verifier.add_keys(key) # this only requires the public key so `verifier.add_keys(key.public_key)` works too
```
```

mickohanlon23: Okay, I added a commit removing calls to all `resolve_keys` class methods, as well as the line mentioning that resolve_keys is supported by `HealthCard` in addition to `HealthCards::Verifier`.

radamson (approved):

PR #106, opened 2023-01-04
Update to Ruby 3.0 with updated dependencies

albertico-gov:
Updating this project to use Ruby 3.0.5. No major Rails update was made.

The following changes were also made in conjunction:

- Updated gems: `rails`, `sqlite3`, `puma`, `sass-rails`, `webpacker`, `turbolinks`, `jbuilder`, `bootsnap`, `fhir_models`, `rubocop-minitest`, `rubocop-rake`, `web-console`, `listen`, `rack-mini-profiler`, `capybara`.
- Updated `Dockerfile` to use Ruby `3.0.5`.
- Updated `Gemfile.lock` files (using bundler `2.4.2`)
- Updated Node.js to LTS version `18.12.1`
- Updated `config/webpacker.yml`, `babel.config.js`, `package.json` and `yarn.lock` using webpacker.
- Compilation of packs via `bin/webpack` required setting `export NODE_OPTIONS=--openssl-legacy-provider` to avoid errors.
- Updated `db/schema.rb` to the version generated by `bin/setup`.
- Modified `README.md` to reflect the Ruby version update.

This solves #105

PR #108, opened 2023-01-17
OpenSSL v3 Support

nscott:
This allows usage of JWKs with OpenSSL v3, which is the default in Ubuntu 22.04. OpenSSL v1.1.1 (the current default in many OSes) is EOL later in 2023.

`rake test` successfully executes with OpenSSL v1.1.1, so this should be fully backwards compatible. None of that code has changed, but instead it's routed to if the OpenSSL version matches.

I haven't been able to make an isolated environment with OpenSSL v3; I think it would be much easier to do once [PR 106](https://github.com/dvci/health_cards/pull/106) is merged.

That said, I *did* pull in this branch to my service which is running Ubuntu 22.04 with OpenSSL v3, and I was able to successfully generate QR codes once again.

The Ruby OpenSSL support for parameters is currently lacking, so using the ASN data to manually create keys seems to be "state of the art". See https://github.com/ruby/openssl/issues/369#issuecomment-818583949 and https://github.com/ruby/openssl/issues/369#issuecomment-1365448588 for specifically using ASNs.

Once https://github.com/ruby/openssl/pull/555 is merged it's likely that a lot of the "ugly" code with ASNs can be cleaned up and made simpler, but this gets everything working in the mean time.


nscott:
The tests failed due to a pdf generation issue on the test platform. Don't think I can help there, unfortunately. Please still review this PR.

ssayer:
Hi @nscott, Thanks for the contribution! Let me poke around with the CI build, it looks like an issue with versioning.

ssayer:
@nscott It looks like the issue is an incompatibility between the version of wicked_pdf in the Gemfile and the Ubuntu version used by the GH CI workflow. PR #106 looks like it used a newer version to solve this problem. If you'd like to try updating the wicked_pdf version in this PR, please do, otherwise we'll look to rebase this once we have the Ruby 3.0 updates in.

nscott:
Just a reminder: https://www.openssl.org/blog/blog/2023/06/15/1.1.1-EOL-Reminder/

